PR #1, opened 2019-08-19
Interface

Jokeswar:


README.md
```diff
@@ -1,2 +1,47 @@
 # interface
 UI for Vmchecker
+[![Build Status](https://jenkins.liquiddemo.org/api/badges/vmck/interface/status.svg)](https://jenkins.liquiddemo.org/vmck/interface)
```
mgax: I like how you created an empty project and started with a PR, props :+1:

interface/minio_api.py
```diff
@@ -0,0 +1,35 @@
+import io
+from minio import Minio
+from django.conf import settings
+
+
+class MinioAPI:
+    __instance = None
+
+    @staticmethod
+    def getInstance():
```
mgax: Ah, the Java singleton pattern :) You could just write `minio_api = MinioAPI()` at the bottom of the file and you have your singleton importable by anyone ;) Even better, get rid of this class, and make `client = Minio(...)` into a module-level object and convert `upload` and `download` to plain functions instead of methods.

interface/settings.py
```diff
@@ -0,0 +1,21 @@
+from .base_settings import * # noqa
```
mgax: I'm starting to think that the pattern of `base_settings` and `settings` is overkill, a single file may be enough. The testsuite can import the main settings file and override whatever it needs.

interface/backend/auth.py
```diff
@@ -0,0 +1,16 @@
+from django.contrib.auth.models import User
+
+class VmckAuthBackend:
```
mgax: Are you sure you need your own authentication backend? We're going to use LDAP anyway, so this code will likely go away. Why not simply use `./manage.py createsuperuser` for now?

server.py
```diff
@@ -0,0 +1,5 @@
+from waitress import serve
+from interface.wsgi import application
+
+if __name__ == '__main__':
```
mgax: Hmmm, could we convert this to a django management command, and do `./manage.py waitress`?

vagrant/Vagrantfile
```diff
@@ -0,0 +1,21 @@
...
+ Vagrant.configure("2") do |config|
+  config.vm.box = "base"
+
+   if Vagrant.has_plugin?('vagrant-env')
+    config.env.enable
+  end
+
+   config.nfs.functional = false
+
+   config.vm.provision 'shell', inline: "cd /vagrant; unzip archive.zip; ls -al"
```
mgax: :+1:

.drone.yml
```diff
@@ -0,0 +1,67 @@
...
+  image: alpine/flake8:3.7.8
+  commands:
+  - flake8
+
+- name: pytest
+  image: vmck/vagrant-vmck:0.2.1
+  commands:
+  - export VMCK_URL=http://$VMCK_IP:$VMCK_PORT
+  - export VMCK_NAME="drone $DRONE_REPO:$DRONE_BRANCH#$DRONE_BUILD_NUMBER"
+  - ./ci/deploy-acs-interface.sh
```
mgax: I think you can move the code from `deploy-acs-interface.sh` to the test itself, it doesn't encapsulate much useful logic. And please move the `Vagrantfile` to the top level of the repo. It's a bad idea to keep it hidden, and it should be reusable by people who want to work on the project, so they don't have to install it locally.

PR #2, opened 2019-09-06
Docs

Jokeswar:
Docs on how to set up a homework


docs/Readme.md
```diff
@@ -0,0 +1,23 @@
+# ACS-INTERFACE DOCUMENTATION
+### True for up to commit: f5d8b0c84838e34dd156974d9407831d363b8900
```
mgax: Yeah that commit thing aint gonna fly. You can use version numbers if you want.

docs/Readme.md
```diff
@@ -0,0 +1,23 @@
...
+
+##Steps to set up a new homework:
+
+1. Create a `.ini` file with the name `{classname}_{homework_number}.ini`
+(E.g. `pc_0.ini` for the first homework) in the directory `{root}/assignment/`.
```
mgax: It would be great if the TA would never log into the vmck or interface server. Let them publish things on github, or the internal gitlab, or wherever. And they can give us a URL when setting up the assignment in the interface.

docs/Readme.md
```diff
@@ -0,0 +1,23 @@
...
+```
+[VMCK]
+image_path = imgbuild-acs-pc.qcow2.tar.gz # image that the vm will use
+memory = 512 # the memory in MB
+cpus = 1 # number of cores, 1 core has 1000 MHZ
+```
+
+2. Add the checker script with the same format name `{classname}_{homework_number}.sh`
+in the same direcotry as above. Must have execution permission and must output the score
+as the last line with the format `{score}/{max_score}`. You can assume that the script will
```
mgax: You should provide an example output.

docs/Readme.md
```diff
@@ -0,0 +1,23 @@
...
+run at the root of the student's submission.
+
+3. NOT IMPLEMENTED YET.
```
mgax: Of course it's not implemented, but we don't need to mention that with all caps in the readme.

PR #3, opened 2019-09-14
Download archives to review

Jokeswar:
It's a django admin action used in the Assignment page. A quick walkthrough: an assistant goes to the `Assignment` page from `django admin`, selects the assignment and gets a tar with the submissions. Now I make some assumptions here:
1) He added his students to a group, and he himself is in that group too.
2) That group is the first one in his user model `groups` list

Is this workflow ok? If so, how can I fix `2`.


interface/admin.py
```diff
@@ -1,6 +1,44 @@
...
+        ta_group = ta_user.groups.all()[0].name
+
+        submissions = Submission.objects.filter(
+                            assignment=assignment,
+                            user__groups__name=ta_group).exclude(user=ta_user)
+
+        with TemporaryDirectory() as _tmp:
+            tmp = Path(_tmp)
+
+            with TarFile(tmp / 'review.tar', 'x') as tar:
```
gabriel-v: Why not just provide the zip instead?

Jokeswar: Use a `zip` archive instead of the `tar` archive? Wasn't the zip in a zip idea bad?

interface/admin.py
```diff
@@ -1,6 +1,44 @@
...
+    actions = ['download_submissions']
+
+    def download_submissions(self, request, _assignment):
+        assignment = _assignment[0]
+        ta_user = request.user
+        ta_group = ta_user.groups.all()[0].name
+
+        submissions = Submission.objects.filter(
+                            assignment=assignment,
+                            user__groups__name=ta_group).exclude(user=ta_user)
```
mgax: Don't worry about excluding the user's own submission. Let the TAs figure this out on their own. Otherwise you get into questions like "what happens if a user is TA for a course and plain student for another course?"

interface/admin.py
mgax: > Wasn't the zip in a zip idea bad?

That was in another context, when you were creating a payload with a payload inside.

> Why not just provide the zip instead?

I think he's adding a bunch of zips, one for each submission? This is basically bulk download.

But for this use case, I'd go with zip, it's easier to work with for the non-geek user.

interface/admin.py
```diff
@@ -1,6 +1,48 @@
...
 admin.site.register(Submission)
-admin.site.register(Assignment)
+
+
+@admin.register(Assignment)
+class AssignmentAdmin(admin.ModelAdmin):
+    actions = ['download_submissions']
+
+    def download_submissions(self, request, _assignment):
+        assignment = _assignment[0]
```
mgax: So you're expecting a single assignment? You should check, and throw an error otherwise.

interface/admin.py
```diff
@@ -1,6 +1,48 @@
...
+
+        subs = Submission.objects.filter(
+                assignment=assignment).order_by(F('timestamp').asc())
+
+        submissions = {}
+
+        for submission in subs:
+            submissions[submission.user.username] = submission
```
mgax: I think you can add a comment like "we only want the last submission", I was confused at first, until I remembered we talked about this.

interface/admin.py
```diff
@@ -1,6 +1,48 @@
...
+            with ZipFile(tmp / 'review.zip', 'x') as zipfile:
+                for _, submission in submissions.items():
+                    submission.download(tmp / f'{submission.id}.zip')
+                    zipfile.write(
+                        tmp / f'{submission.id}.zip',
+                        f'{submission.user.username}-{submission.id}.zip')
+
+            with open(tmp / 'review.zip', 'rb') as zipfile:
+                response = HttpResponse(zipfile.read(),
+                                        content_type='application/zip')
```
mgax: https://docs.djangoproject.com/en/2.2/ref/request-response/#fileresponse-objects

It expects an open file from which it will stream the content. Your zip file will be unlinked when you exit the `with TemporaryDirectory` block, but thats' ok, you can open a filehandle on `tmp / 'review.zip'`, pass it on to `FileResponse`, and the actual file on disk will be removed once that filehandle is closed.

```python
with TemporaryDirectory() as _tmp:
    # ...
    review_zip = (tmp / 'review.zip').open('rb')
# TemporaryDirectory deletes itself and the `review.zip` hard link,
# but the kernel knows, you have the file open, so it doesn't
# delete the inode quite yet.
return FileResponse(review_zip)
```

interface/admin.py
```diff
@@ -1,6 +1,49 @@
...
 admin.site.register(Submission)
-admin.site.register(Assignment)
+
+
+@admin.register(Assignment)
+class AssignmentAdmin(admin.ModelAdmin):
+    actions = ['download_submissions']
+
+    def download_submissions(self, request, _assignment):
+        if _assignment.count() != 1:
```
mgax: Prefixing an argument with `_` usually means it's ignored/irrelevant. You can name this one `queryset`, it's a good convention for admin actions.

interface/admin.py
```diff
@@ -1,6 +1,49 @@
...
+            raise RuntimeError('Only one assignment can be selected')
```
mgax: Ah, but if the app runs with `DEBUG = False`, the user will see a 500 error with no explanation. The polite thing to do is to set a [message](https://docs.djangoproject.com/en/2.2/ref/contrib/messages/) and simply return (I believe Django will redirect the user back to the admin page where they came from).

interface/admin.py
```diff
@@ -1,6 +1,49 @@
...
+
+        assignment = _assignment[0]
+
+        subs = Submission.objects.filter(
+                assignment=assignment).order_by(F('timestamp').asc())
```
mgax: I think you can simply write `.order_by('timestamp')`, no? And the `assignment` object should have a `submission_set` attribute which is already the filtered queryset you want, except for order_by.

interface/admin.py
```diff
@@ -1,6 +1,49 @@
...
+
+        # we only want the last submission of every user
+        for submission in subs:
+            submissions[submission.user.username] = submission
+
+        with TemporaryDirectory() as _tmp:
+            tmp = Path(_tmp)
+
+            with ZipFile(tmp / 'review.zip', 'x') as zipfile:
+                for _, submission in submissions.items():
```
mgax: Ah, you can do `for submission in submissions.values()` ;)

interface/admin.py
```diff
@@ -1,6 +1,49 @@
...
+                    submission.download(tmp / f'{submission.id}.zip')
+                    zipfile.write(
+                        tmp / f'{submission.id}.zip',
+                        f'{submission.user.username}-{submission.id}.zip')
```
mgax: You can close the bracket on the next line for symmetry :)

interface/admin.py
```diff
@@ -1,6 +1,49 @@
...
+
+            review_zip = (tmp / 'review.zip').open('rb')
+            response = FileResponse(review_zip)
+
+            return response
```
mgax: Why not simply `return FileResponse(...)`?

interface/admin.py
```diff
@@ -1,6 +1,46 @@
...
+
+
+@admin.register(Assignment)
+class AssignmentAdmin(admin.ModelAdmin):
+    actions = ['download_submissions']
+
+    def download_submissions(self, request, queryset):
+        if queryset.count() != 1:
+            messages.warning(request, 'Only one assignment can be selected')
+            return request
```
mgax: It seems weird to return a request, when the function returns responses, are you sure this is right? Maybe just `return`?

interface/admin.py
```diff
@@ -1,6 +1,46 @@
...
+        with TemporaryDirectory() as _tmp:
+            tmp = Path(_tmp)
+
+            with ZipFile(tmp / 'review.zip', 'x') as zipfile:
+                for submission in submissions.values():
+                    submission.download(tmp / f'{submission.id}.zip')
+                    zipfile.write(
+                        tmp / f'{submission.id}.zip',
+                        f'{submission.user.username}-{submission.id}.zip'
+                        )
```
mgax: Please unindent the closing bracket. Flake8 says `41 col 25| E123 closing bracket does not match indentation of opening bracket's line`. Also, add a final comma to the last argument of the function.

interface/admin.py
Jokeswar: It works either way. I'll leave `return` then.

interface/admin.py
Jokeswar: It's strange because the CI didn't cry about it.

interface/models.py
```diff
@@ -30,6 +30,10 @@ class Assignment(models.Model):
     def full_code(self):
         return f'{self.course.code}-{self.code}'
 
+    @property
+    def submission_set(self):
```
mgax: Wait, `submission_set` is already defined by Django, because of the `ForeignKey` from `Submission` to `Assignment`. You're duplicating Django's magic.

interface/admin.py
```diff
@@ -1,6 +1,46 @@
...
-admin.site.register(Assignment)
+
+
+@admin.register(Assignment)
+class AssignmentAdmin(admin.ModelAdmin):
+    actions = ['download_submissions']
+
+    def download_submissions(self, request, queryset):
+        if queryset.count() != 1:
+            messages.warning(request, 'Only one assignment can be selected')
```
mgax: `error` instead of `warning`, because the requested operation failed.

interface/models.py
Jokeswar: I don't get it. Explain it in more depth please. Given an `Assignment` i need to get all of it's corresponding `Submissions`. How should it look like?

mgax: https://docs.djangoproject.com/en/2.2/ref/models/relations/ <- Django creates a `submission_set` RelatedManager on any `Assignment` instance.

PR #12, opened 2019-10-01
Quick fixes

Jokeswar:
Readme and whitelist

### DIFF
--- a/deploy/interface.nomad
+++ b/deploy/interface.nomad
@@ -23,9 +23,11 @@ job "acs-interface" {
       }
       template {
         data = <<-EOF
-          MINIO_ACCESS_KEY = "1234"
-          MINIO_SECRET_KEY = "123456789"
-          MINIO_BROWSER = "on"
+          {{- with secret "kv/minio" -}}
+            MINIO_ACCESS_KEY = "{{ .Data.access_key }}"
+            MINIO_SECRET_KEY = "{{ .Data.secret_key }}"
+            MINIO_BROWSER = "on"
+          {{- end -}}
           EOF
           destination = "local/config.env"
           env = true
--- a/deploy/interface.nomad
+++ b/deploy/interface.nomad
@@ -26,11 +26,11 @@ job "acs-interface" {
           {{- with secret "kv/minio" -}}
             MINIO_ACCESS_KEY = "{{ .Data.access_key }}"
             MINIO_SECRET_KEY = "{{ .Data.secret_key }}"
-            MINIO_BROWSER = "on"
           {{- end -}}
-          EOF
-          destination = "local/config.env"
-          env = true
+            MINIO_BROWSER = "on"
+        EOF
+        destination = "local/config.env"
+        env = true
       }
       resources {
         memory = 200
--- a/deploy/interface.nomad
+++ b/deploy/interface.nomad
@@ -142,6 +142,10 @@ job "acs-interface" {
       service {
         name = "acs-interface"
         port = "http"
+        tags = [
+          "traefik.enable=true",
+          "traefik.frontend.rule=Host:vmchecker.liquiddemo.org",
+        ]
         check {
           name = "acs-interface alive on http"
           initial_status = "critical"
--- a/deploy/interface.nomad
+++ b/deploy/interface.nomad
@@ -61,7 +61,7 @@ job "acs-interface" {
       }
       driver = "docker"
       config {
-        image = "vmck/acs-interface:interface"
+        image = "vmck/acs-interface"
         dns_servers = ["${attr.unique.network.ip-address}"]
         command = "/bin/bash"
         args = [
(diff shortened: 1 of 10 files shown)


interface/settings.py
```diff
@@ -108,6 +108,7 @@
 MINIO_BUCKET = os.environ.get('MINIO_BUCKET', 'test')
 
 ACS_INTERFACE_ADDRESS = os.environ.get('ACS_INTERFACE_ADDRESS', 'localhost:8100')  # noqa: E501
+ACS_USER_WHITELIST = os.getenv('ACS_USER_WHITELIST', ',').split(',')
```
mgax: `','` makes no sense as a default, and even `''` will result in the split value `['']`. The clean way is to use an intermediate variable:

```python
_acs_user_whitelist = os.getenv('ACS_USER_WHITELIST', None)
ACS_USER_WHITELIST = _acs_user_whitelist.split(',') if _acs_user_whitelist else []

File: deploy/interface.nomad [hunk 2]
```diff
@@ -23,9 +23,11 @@ job "acs-interface" {
       }
       template {
         data = <<-EOF
-          MINIO_ACCESS_KEY = "1234"
-          MINIO_SECRET_KEY = "123456789"
-          MINIO_BROWSER = "on"
+          {{- with secret "kv/minio" -}}
+            MINIO_ACCESS_KEY = "{{ .Data.access_key }}"
+            MINIO_SECRET_KEY = "{{ .Data.secret_key }}"
+            MINIO_BROWSER = "on"
```
mgax: You can move this line outside the `with secret` block.

deploy/interface.nomad
```diff
@@ -23,9 +23,11 @@ job "acs-interface" {
...
+          {{- end -}}
           EOF
           destination = "local/config.env"
           env = true
```
mgax: It would make sense to unindent these two lines, `EOF` is one tab extra.

interface/admin.py
```diff
@@ -44,3 +45,24 @@ def download_submissions(self, request, queryset):
...
     download_submissions.short_description = 'Download submissions for review'
+
+
+@admin.register(Submission)
+class SubmissionAdmin(admin.ModelAdmin):
+    actions = ['rerun_submissions']
+
+    def rerun_submissions(self, request, submissions):
+        for old_submission in submissions:
+            new_submission = Submission.objects.create(
```
mgax: Ah, you're creating new submissions, but I think that's overkill, and counterproductive, because a submission maps to a user action (please take this zip and evaluate it), makes no sense to create another one. Just re-deploy it.

interface/backend/submission.py
```diff
@@ -42,26 +42,7 @@ def vmck_config(submission):
...
-
-    submission = Submission.objects.create(
-        archive_size=file.size,
-        user=request.user,
-        assignment=assignment,
-    )
-
-    storage.upload(f'{submission.id}.zip', file.read())
-
+def deploy_submission(submission, assignment):
```
mgax: Yes, extract this, nice. But `deploy` is a concept that makes sense in a cluster (nomad), whereas here we're more like evaluating, so how about `evaluate_submission`? Also you can make it a method of the `submission` object, so it becomes `submission.evaluate()`.

interface/backend/submission.py
```diff
@@ -91,6 +72,29 @@ def handle_submission(request):
 
     submission.vmck_job_id = response.json()['id']
```
mgax: You can call `submission.save()` here because you're updating the record.

interface/views.py
```diff
@@ -114,12 +115,15 @@ def done(request, pk):
                                    pk=pk,
                                    score__isnull=True)
 
-    stdout = utils.decode(options['stdout']).split('\n')
+    stdout = utils.decode(options['stdout'])
```
mgax: :+1:

interface/views.py
```diff
@@ -114,12 +115,15 @@ def done(request, pk):
...
     stderr = utils.decode(options['stderr'])
     exit_code = int(options['exit_code'])
 
-    submission.score = int(stdout[-2].split('/')[0])
-    submission.output = '\n'.join(stdout[:-2])
+    score = re.search(r'.*TOTAL: (\d+)/(\d+)', stdout, re.MULTILINE)
+    points = score.group(1) if score else 0
```
mgax: If score is `None` you can even log a warning here.

interface/views.py
```diff
@@ -114,12 +115,15 @@ def done(request, pk):
...
+
+    submission.score = points
+    submission.output = stdout + '\n' + stderr
```
mgax: Eh, this is not quite correct, you should either capture stdout and stderr together so they interleave messages, or save them as separate fields. Concatenating them is impolite. If this is too much hassle for now, ok, but make an issue.

Jokeswar: issue #14

interface/utils.py
```diff
@@ -1,10 +1,39 @@
...
+    m = re.match(r'https://github.com/(?P<org>[^/]+)/(?P<repo>[^/]+)/?$',
+                 submission.assignment.repo_url)
+    url_base = ('https://raw.githubusercontent.com'
+                '/{0}/{1}/'.format(*list(m.groups())))
+
+    config_data = requests.get(
+                    urljoin(
+                        url_base,
+                        f'{submission.assignment.repo_branch}/config.ini')
+                        )
```
mgax: That's some wacky indentation bro. A true artist. :p

Issue #15, opened 2019-10-02
Review pop-up


Jokeswar:
TA clicks a button on a submission report, a textarea opens, and they type in vmchecker style grading.

mgax:
@murarugeorgec could you please provide an example of vmr grading for us?

murarugeorgec:
Sure. It looks something like this:
```
-1.0: Indentare neconsecventa
-2.5: Functii si variabile cu nume nesugestive
    Ex: a, b, functie1
```
Cateodata lumea mai trece la finalul fisierului cine si cand a corectat. Ex
```
George Muraru, 17.01.2018
```
Exemplul este creat pe moment - insa asa arata un fisier vmr grade.
Initial (inainte de corectare) acesta are in el textul ```ok```

PR #21, opened 2019-10-02
Pop up

Jokeswar:
#15


interface/templates/interface/review_modal.html
```diff
@@ -0,0 +1,39 @@
...
+<!-- Data in modal -->
+<script>
+    $('#review').on('show.bs.modal', function (event)
+    {
+        var button = $(event.relatedTarget)
+        var recipient = button.data('id')
+
+        var modal = $(this)
+        modal.find(".modal-body form")[0].action = "{{ sub_base_url }}" + recipient + "/review"
+        modal.find(".modal-title")[0].innerText = "Review submission " + recipient
```
mgax: Ah, front-end hacks at their finest. You are on the right track. :)

interface/templates/interface/submission_list.html
```diff
@@ -8,6 +8,12 @@
         <!-- Bootstrap CSS -->
         <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
 
+        <!-- Holly Javascript -->
+        <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN" crossorigin="anonymous"></script>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.12.9/umd/popper.min.js" integrity="sha384-ApNbgh9B+Y1QKtv3Rn7W3mgPxhU9K/ScQsAP7hUibX39j7fakFPskvXusvfa0b4Q" crossorigin="anonymous"></script>
+        <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/js/bootstrap.min.js" integrity="sha384-JZR6Spejh4U02d8jOt6vLEHfe/JQGiRRSQQxSfFWpi1MquVdAyjUar5+76PVCmYl" crossorigin="anonymous"></script>
```
mgax: Pick one CDN and stick to it. My favourite these days is [unpkg](https://unpkg.com/), they make short and readable URLs.

interface/views.py
```diff
@@ -74,6 +74,15 @@ def homepage(request):
                    'logout_url': redirect(logout_view).url})
 
 
+@login_required
+@csrf_exempt
```
mgax: You sure this is ok, or even required? Have a read through https://docs.djangoproject.com/en/2.2/ref/csrf/ - in your case I think you can use a hidden input in the form.

interface/views.py
```diff
@@ -74,6 +74,24 @@ def homepage(request):
                    'logout_url': redirect(logout_view).url})
 
 
+@login_required
+def review(request, pk):
+    if request.POST and request.user.is_staff:
+        marks = re.findall('^(\+|-[0-9]+\.[0-9]+):',  # noqa: W605
```
mgax: Can we not add `noqa` if we can avoid it? Those warnings are there for a reason.

Also, try this indent, I think it's nicer:

```python
marks = re.findall(
    '^(\+|-[0-9]+\.[0-9]+):',
    request.POST['review-code'],
    re.MULTILINE,
)
```

Also, Python has `r''` ([raw strings](https://docs.python.org/3/library/re.html#raw-string-notation)) for regular expressions, which treat escapes differently. In other words, your backslash escapes don't do what you expect them to do.

interface/models.py
```diff
@@ -72,13 +72,16 @@ class Submission(models.Model):
                                    null=True)
     user = models.ForeignKey(User, on_delete=models.PROTECT, null=True)
     output = models.CharField(max_length=4096, default='none')
+    review_message = models.CharField(max_length=1024, default='none')
```
mgax: Let's be more generous for the talkative TA, it costs us nothing to give them more space :)

interface/views.py
```diff
@@ -74,6 +75,26 @@ def homepage(request):
...
+        marks = re.findall(
+            r'^([+-]\d\.\d):',
+            request.POST['review-code'],
+            re.MULTILINE
+        )
+        log.debug('Marks found: ' + str(marks))
+
+        review_score = sum([decimal.Decimal(mark) for mark in marks])
+
+        submission = get_object_or_404(models.Submission, pk=pk)
```
mgax: You should do this at the beginning of the function, if `pk` is invalid then 404 is the only answer.

interface/templates/interface/submission_result.html
```diff
@@ -6,19 +6,35 @@
...
-        <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+        <link rel="stylesheet" href="https://unpkg.com/bootstrap@4.3.1/dist/css/bootstrap.min.css">
+
+        <!-- Holly Javascript -->
+        <script src="http://unpkg.com/jquery@3.2.1/dist/jquery.js"></script>
+        <script src="https://unpkg.com/bootstrap@4.3.1/dist/js/bootstrap.js"></script>
 
         <title>VMCK</title>
+
+        {% include "interface/review_modal.html" %}
```
mgax: I'm pretty sure this is not supposed to be within `<head>`. A better place would be at the very bottom of the page, before `</body>`.

interface/templates/interface/submission_result.html
```diff
@@ -6,19 +6,35 @@
...
-            <a class="btn btn-primary" href='{{ submission_list_url }}'>List submissions</a>
-            <p>Submission id: {{ sub.id }}</p>
-            <p>Your score is: {{ sub.score }}/{{ sub.assignment.max_score }}</p>
-            <p>The output is:</p>
-            <pre><samp>{{ sub.output }}</samp></pre>
+            <div class="row mb-2 mt-2">
+                <a class="btn btn-primary mr-1" href='{{ homepage_url }}'>Homepage</a>
+                <a class="btn btn-primary" href='{{ submission_list_url }}'>List submissions</a>
+            </div>
+            <div class="container">
```
mgax: A `container` within a `container` looks suspicious. I think it's basically a no-op, but it could confuse the layout.

interface/views.py
```diff
@@ -74,6 +75,26 @@ def homepage(request):
                    'logout_url': redirect(logout_view).url})
 
 
+@login_required
+def review(request, pk):
+    if request.POST and request.user.is_staff:
```
mgax: You should check for `is_staff` at the top of the view, using the [staff_member_required](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.views.decorators.staff_member_required) decorator. Also, the view should only take `POST` requests, and you can do this with the [require_http_methods](https://docs.djangoproject.com/en/2.2/topics/http/decorators/#django.views.decorators.http.require_http_methods) decorator.

interface/views.py
```diff
@@ -74,6 +75,26 @@ def homepage(request):
...
+        marks = re.findall(
+            r'^([+-]\d\.\d):',
+            request.POST['review-code'],
+            re.MULTILINE
```
mgax: Final comma please :/

interface/views.py
```diff
@@ -74,6 +77,28 @@ def homepage(request):
                    'logout_url': redirect(logout_view).url})
 
 
+@login_required
+@staff_member_required
+@require_POST
+def review(request, pk):
+    submission = get_object_or_404(models.Submission, pk=pk)
```
mgax: 👌

Issue #22, opened 2019-10-06
Move regular expression separately


Jokeswar:
This piece of [code](https://github.com/vmck/acs-interface/blob/master/interface/models.py#L117-L125) appears [here](https://github.com/vmck/acs-interface/blob/master/interface/utils.py#L14-L25) too. Move it into a separate function with some better var names.

izhuravlev:
Hey @Jokeswar,
I can do that. Are there any other recommendations on this issue?

izhuravlev:
I have also mentioned, that in function ```evaluate(self)```  from **models.py**  [this](https://github.com/vmck/acs-interface/blob/b98ca09d2146cf08c346ba833a78f3f33f960224/interface/models.py#L117-L125) piece of code is being called right before calling ```vmck_config(self)```, defined in the **utils.py**. ```vmck_config``` function has the same block of code in it - so do we actually need it in the ```evaluate(self)``` function?

izhuravlev:
Okay, I've got the difference and the importance of this piece of code. 
Later in the **models.py** file (line 132), we need the ```config_url``` variable - the result from line 123. 
As I have also mentioned, in **models.py** we need only the URL, while in **utils.py** we need the data - so I believe we have to create 2 separate functions - something like _getConfigUrl_ and _getConfigData_. 
I also propose to create a third function _getUrlBase_ with functionality from **utils.py** lines 14-18 and call it in functions _getConfigUrl_ and _getConfigData_.

izhuravlev:
I will be happy to make any corrections if they are required.

PR #23, opened 2019-10-06
issue 22 fix

izhuravlev:
Fixes: #22  
I've added ```get_url_base```, ```get_config_url``` and ```get_config_data``` functions. 
For more information, see issue #22 [comments](https://github.com/vmck/acs-interface/issues/22#issuecomment-538797682)

### DIFF
--- a/interface/models.py
+++ b/interface/models.py
@@ -114,14 +114,7 @@ class Submission(models.Model):
         return self.STATE_CHOICES[self.state]
 
     def evaluate(self):
-        m = re.match(r'https://github.com/(?P<org>[^/]+)/(?P<repo>[^/]+)/?$',
-                     self.assignment.repo_url)
-
-        url_base = ('https://raw.githubusercontent.com/'
-                    '{0}/{1}/'.format(*list(m.groups())))
-
-        config_url = urljoin(url_base,
-                             f'{self.assignment.repo_branch}/checker.sh')
+        config_url = get_config_url(self)
 
         options = vmck_config(self)
         options['name'] = f'{self.assignment.full_code} submission #{self.id}'
--- a/interface/models.py
+++ b/interface/models.py
@@ -114,7 +114,7 @@ class Submission(models.Model):
         return self.STATE_CHOICES[self.state]
 
     def evaluate(self):
-        config_url = get_config_url(self)
+        script_url = get_script_url(self)
 
         options = vmck_config(self)
         options['name'] = f'{self.assignment.full_code} submission #{self.id}'
@@ -122,7 +122,7 @@ class Submission(models.Model):
         options['env'] = {}
         options['env']['archive'] = self.get_url()
         options['env']['vagrant_tag'] = settings.MANAGER_TAG
-        options['env']['script'] = config_url
+        options['env']['script'] = script_url
         options['env']['memory'] = settings.MANAGER_MEMORY
         options['env']['cpu_mhz'] = settings.MANAGER_MHZ
         options['env']['callback'] = urljoin(
--- a/interface/utils.py
+++ b/interface/utils.py
@@ -11,17 +11,7 @@ vocabulary_64 = string.ascii_letters + string.digits + '.+'
 
 
 def vmck_config(submission):
-    m = re.match(r'https://github.com/(?P<org>[^/]+)/(?P<repo>[^/]+)/?$',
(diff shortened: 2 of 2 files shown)


interface/utils.py
```diff
@@ -57,3 +47,30 @@ def random_code(length, vocabulary=vocabulary_64):
 def is_true(value):
     text = (value or '').lower().strip()
     return text in ['1', 'yes', 'true', 'on', 'enabled']
+
+    
+def get_url_base(link):
+    m = re.match(r'https://github.com/(?P<org>[^/]+)/(?P<repo>[^/]+)/?$',
+                    link.assignment.repo_url)
```
Jokeswar: The `link.assignment.repo_url` should start right below `r`. Something like:
```
m = re.match(r'https://github.com/(?P<org>[^/]+)/(?P<repo>[^/]+)/?$',
             link.assignment.repo_url)
```

I suggest using `flake8` as linter. We want the code to respect the `pep8` standard.

interface/utils.py
```diff
@@ -57,3 +47,30 @@ def random_code(length, vocabulary=vocabulary_64):
...
+    url_base = ('https://raw.githubusercontent.com/'
+                '{0}/{1}/'.format(*list(m.groups())))
+    return url_base
+
+def get_config_url(link):
```
Jokeswar: I think `get_script_url` would be more appropriate.

interface/models.py
```diff
@@ -114,14 +114,7 @@ def state_label(self):
...
     def evaluate(self):
-        m = re.match(r'https://github.com/(?P<org>[^/]+)/(?P<repo>[^/]+)/?$',
-                     self.assignment.repo_url)
-
-        url_base = ('https://raw.githubusercontent.com/'
-                    '{0}/{1}/'.format(*list(m.groups())))
-
-        config_url = urljoin(url_base,
-                             f'{self.assignment.repo_branch}/checker.sh')
+        config_url = get_config_url(self)
```
Jokeswar: `script_url = get_script_url(self)` would be better. Please do not forget to update it a few lines bellow too, at `options['env']['script']`

interface/utils.py
```diff
@@ -57,3 +47,30 @@ def random_code(length, vocabulary=vocabulary_64):
...
+    config_data = requests.get(
+                    urljoin(
+                        url_base,
+                        f'{link.assignment.repo_branch}/config.ini',
+                    )
+                  )
+    return config_data
+
+
+
```
Jokeswar: Too many new lines at the end of the file. Should be just one, right after `return config_data`. Also, a new line just before `return` would look really nice.

interface/utils.py
```diff
@@ -57,3 +47,30 @@ def random_code(length, vocabulary=vocabulary_64):
...
+                    link.assignment.repo_url)
+    url_base = ('https://raw.githubusercontent.com/'
+                '{0}/{1}/'.format(*list(m.groups())))
+    return url_base
+
+def get_config_url(link):
+    url_base = get_url_base(link)
+    config_url = urljoin(url_base,
+                             f'{link.assignment.repo_branch}/checker.sh')
+    return config_url
```
Jokeswar: Same thing about the coding style as above applies here. Also, `script_url` is more intuitive than `config_url`.

interface/utils.py
Jokeswar: Two new lines between functions.

interface/utils.py
```diff
@@ -57,3 +47,30 @@ def random_code(length, vocabulary=vocabulary_64):
...
+
+def get_config_data(link):
```
Jokeswar: Two new lines between functions.

izhuravlev:
Sure, will fix it right away!

izhuravlev:
@Jokeswar made requested corrections

interface/utils.py
```diff
@@ -57,3 +47,32 @@ def random_code(length, vocabulary=vocabulary_64):
...
+    url_base = get_url_base(link)
+    config_data = requests.get(
+                    urljoin(
+                        url_base,
+                        f'{link.assignment.repo_branch}/config.ini',
+                    )
+                  )
+
+    return config_data
```
Jokeswar: New line at the end of the file. Something like:
```
return config_data

```

PR #25, opened 2019-10-07
Use jwt to authenticate

Jokeswar:
Uses vmck/vmck#105 and vmck/vagrant-vmck#31


interface/models.py
```diff
@@ -132,6 +133,9 @@ def evaluate(self):
         options['env']['script'] = config_url
         options['env']['memory'] = settings.MANAGER_MEMORY
         options['env']['cpu_mhz'] = settings.MANAGER_MHZ
+        options['env']['auth'] = jwt.encode(str(self.id),
+                                            settings.SECRET_KEY,
+                                            algorithms=['HS256'])
```
mgax: I kind of get what you're doing here, but it's pretty opaque. I suggest you move this line and its decode pair to functions in a separate module, with docstrings that explain the logic.

PR #39, opened 2019-10-11
Using postgres as db backend

Jokeswar:
Moving away from `sqlite3`. Solves #39


interface/templates/interface/review_modal.html
```diff
@@ -36,5 +36,6 @@ <h5 class="modal-title" id="reviewLabel"> </h5>
         var modal = $(this)
         modal.find(".modal-body form")[0].action = "{{ sub_base_url }}" + recipient + "/review"
         modal.find(".modal-title")[0].innerText = "Review submission " + recipient
+        modal.find(".modal-body textarea")[0].value = `{{ submission_review_message }}`
```
mgax: This is ripe for cross-site scripting. Good thing only TAs can write here.

PR #42, opened 2019-10-12
removed checking in ACS_USER_WHITELIST

lizababu:
removed checking in ACS_USER_WHITELIST

### DIFF
--- a/interface/views.py
+++ b/interface/views.py
@@ -32,7 +32,7 @@ def login_view(request):
         if form.is_valid():
             user = authenticate(username=form.data['username'],
                                 password=form.data['password'])
-            if user and user.username in settings.ACS_USER_WHITELIST:
+            if user and user.username:
                 login(request, user)
                 return redirect(homepage)
     else:


interface/views.py
```diff
@@ -32,7 +32,7 @@ def login_view(request):
         if form.is_valid():
             user = authenticate(username=form.data['username'],
                                 password=form.data['password'])
-            if user and user.username in settings.ACS_USER_WHITELIST:
```
murarugeorgec: Q: @Jokeswar shouldn't we remove the ```ACS_USER_WHITELIST``` from settings?

lizababu: I removed it from settings since it's not used anywhere else (checked with @Jokeswar)

interface/views.py
```diff
@@ -32,7 +32,7 @@ def login_view(request):
...
+            if user and user.username:
```
Jokeswar: I believe it should just be `if user:`. Before we were checking if the user authenticated AND if it was allowed to enter the site.

PR #45, opened 2019-10-12
Capture stdout and stderr in separate fields#14

stefansava21:
Added a new char field for stderr


interface/models.py
```diff
@@ -72,6 +72,7 @@ class Submission(models.Model):
                                    null=True)
     user = models.ForeignKey(User, on_delete=models.PROTECT, null=True)
     output = models.CharField(max_length=4096, default='none')
+    output_err = models.CharField(max_length=4096, default='none')
```
Jokeswar: What do you say about naming these 2 guys `stdout` and `stderr`? Previously `output` was fine because we concatenated everything in there. Also you will have to do `pipenv run ./manage.py makemigrations` again to update the database changes. To not get another migrations file(one migrations file per PR is what we are aiming for), please delete `interface/migrations/0005_submission_output_err.py`

interface/views.py
```diff
@@ -153,10 +153,11 @@ def done(request, pk):
         log.warning('Score is None')
 
     submission.score = points
-    submission.output = stdout + '\n' + stderr
+    submission.output = stdout
+    submission.output_err = stderr
```
Jokeswar: Don't forget to rename these.

interface/views.py
```diff
@@ -153,10 +153,11 @@ def done(request, pk):
         log.warning('Score is None')
```
Jokeswar: To test this part of the code:
* comment out `assert submission.verify_jwt(request.GET.get('token'))`
* Change `stdout` and `stderr` to
```
stdout = options['stdout']
stderr = options['stderr']
```
* have at least one submission in the database

Make a POST request(maybe `curl`) to `http://localhost:8100/submission/1/done` with the following JSON:
```
{
"stdout": "what stdout you want! Must have TOTAL: 170/170",
"stderr": "error maybe",
"exit_code": 0
}
```
Now you can check in the `admin` tab, under `submissions` if the submission with the id 1 has the correct `stdout` and `stderr` values.

interface/views.py
```diff
@@ -149,19 +149,18 @@ def submission_list(request):
 @login_required
 def submission_result(request, pk):
     sub = get_object_or_404(Submission, pk=pk)
+    show_review_message = True if not sub.review_message.startswith('none') else False  # noqa: E501
```
mgax: This is silly. The default for `review_message` should be an empty string and it should have `allow_empty=True`. `none` is not a great default.

https://github.com/vmck/acs-interface/blob/master/interface/models.py#L75

interface/views.py
```diff
@@ -181,18 +180,20 @@ def done(request, pk):
...
     submission.score = decimal.Decimal(points)
     submission.total_score = submission.calculate_total_score()
-    submission.output = output
+    submission.stdout = stdout
+    submission.stderr = stderr
     submission.state = Submission.STATE_DONE
 
-    log.debug(f'Submission #{submission.id} has the output:\n{submission.output}')  # noqa: E501
-    log.debug(f'Stderr:\n{stderr}')
+    log.debug(f'Submission #{submission.id} has the stdout:\n{submission.stdout}')  # noqa: E501
```
mgax: Please split this line and print the "stdout" part like you are printing "stderr" below, then it won't be over 79 chars.

interface/models.py
```diff
@@ -50,7 +50,7 @@ class Submission(models.Model):
     username -- the user id provided by the LDAP
     assignment_id -- class specific, will have the form
                      `{course_name}_{homework_id}` for example pc_00
-    message -- the output message of the checker
+    stdout -- the output message of the checker
```
mgax: And `stderr` too please :)

interface/models.py
```diff
@@ -71,7 +71,8 @@ class Submission(models.Model):
                                    on_delete=models.PROTECT,
                                    null=True)
     user = models.ForeignKey(User, on_delete=models.PROTECT, null=True)
-    output = models.CharField(max_length=32768, default='none')
+    stdout = models.CharField(max_length=32768, default='none')
+    stderr = models.CharField(max_length=32768, default='none')
```
mgax: Please default to an empty string.

interface/migrations/0007_auto_20191025_1756.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('interface', '0006_auto_20191016_1355'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
mgax: I suggest we do a `RenameField` so we don't lose data.

interface/templates/interface/submission_result.html
```diff
@@ -40,7 +40,8 @@ <h2>Review</h2>
     <h2>Output</h2>
     <div class="card">
       <div class="card-body">
-        <pre><code>{{ sub.output }}</code></pre>
+          <pre><code>{{ sub.stderr }}</code></pre>
```
mgax: A couple extra spaces indentation here, pls remove.

PR #46, opened 2019-10-12
Added Download All Submissions Option

stefanalex99:
Is the downloaded archive format ok?


interface/admin.py
```diff
@@ -45,6 +50,26 @@ def download_submissions(self, request, queryset):
...
+
+            with ZipFile(tmp / 'review.zip', 'x') as zipfile:
+                for submission in submission_set:
+                    submission.download(tmp / f'{submission.id}.zip')
+                    zipfile.write(
+                        tmp / f'{submission.id}.zip',
+                        f'{submission.user.username}-{submission.id}.zip',
+                    )
+
+            review_zip = (tmp / 'review.zip').open('rb')
```
Jokeswar: This code overlaps with the one above. This bit should be moved in a separate function, something like `archive_submissions(submissions)` that will return a `File object` (i.e. `review_zip`).

interface/admin.py
```diff
@@ -13,16 +13,21 @@
...
         if queryset.count() != 1:
             messages.error(request, 'Only one assignment can be selected')
             return
 
         assignment = queryset[0]
         submission_set = assignment.submission_set.order_by('timestamp')
 
+        return submission_set
+
+    def download_submissions(self, request, queryset):
```
Jokeswar: Name it better. Something like `download_review_submissions` should do it. Also don't forget to rename the part at the end of the function too.

PR #47, opened 2019-10-12
Stay logged in

stefanalex99:


interface/views.py
```diff
@@ -27,6 +27,9 @@
 
 
 def login_view(request):
+    if (request.user.is_authenticated):
```
Jokeswar: No parenthesis please.

interface/admin.py
```diff
@@ -13,16 +13,21 @@
 
 @admin.register(Assignment)
 class AssignmentAdmin(admin.ModelAdmin):
-    actions = ['download_submissions']
+    actions = ['download_submissions', 'download_all_submissions']
```
Jokeswar: This part is the same as #46. I suggest using `git rebase --onto master origin/downloadAllSubmissions  StayLoggedIn` to keep only the logged in part. You will get conflicts when we will try to merge this branch after #46, or the other way around.

stefanalex99: I am not sure if I solved the problem, can you check?

Jokeswar: It didn't solve it. You can check here to see the differences: https://github.com/vmck/acs-interface/pull/47/files

PR #48, opened 2019-10-12
added total_score field that is updated at review and submission#38

lizababu:
The correct version of the code.


interface/models.py
```diff
@@ -89,7 +92,7 @@ def get_url(self):
         return storage.get_link(f'{self.id}.zip')
 
     @property
```
Jokeswar: You should remove this line. This is a fully fledged function, not a property. Plus by leaving `@property` it isn't callable anymore.

interface/views.py
```diff
@@ -153,6 +154,7 @@ def done(request, pk):
         log.warning('Score is None')
 
     submission.score = points
```
Jokeswar: To be sure use `int(points)`, even though `score` is a `IntegerField`, python might still do something funny.

PR #63, opened 2019-10-16
Final tweaks

Jokeswar:


interface/admin.py
```diff
@@ -13,27 +13,20 @@
...
 class AssignmentAdmin(admin.ModelAdmin):
-    actions = ['download_submissions']
-
-    def download_submissions(self, request, queryset):
-        if queryset.count() != 1:
-            messages.error(request, 'Only one assignment can be selected')
-            return
+    actions = ['download_review_submissions', 'download_all_submissions']
 
+    def get_submission_set(self, request, queryset):
```
mgax: So this function simply returns `submission_set.order_by('timestamp')`? Why not remove it and move the code inline?

interface/admin.py
```diff
@@ -13,27 +13,20 @@
...
         submission_set = assignment.submission_set.order_by('timestamp')
 
-        submissions = {}
-
-        # we only want the last submission of every user
-        for submission in submission_set:
-            submissions[submission.user.username] = submission
+        return submission_set
 
+    def archive_submissions(self, submissions):
```
mgax: How about `zip_response_with_submissions`? The name `archive_submissions` means "take these submissions and archive them somewhere".

Issue #74, opened 2019-10-22
log in fails when ldap is not configured


MadalinaCaea:
```
initialize() argument 1 must be str, not None 
/home/vagrant/.local/share/virtualenvs/vagrant-gKDsaKU3/lib/python3.6/site-packages/ldap/functions.py in _ldap_function_call, line 55
```

PR #75, opened 2019-10-22
check if ldap is configured

MadalinaCaea:
fixes https://github.com/vmck/acs-interface/issues/74

### DIFF
--- a/interface/settings.py
+++ b/interface/settings.py
@@ -55,7 +55,7 @@ AUTHENTICATION_BACKENDS = [
 ]
 
 if os.environ.get('LDAP_SERVER_URI'):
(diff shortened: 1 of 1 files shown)


interface/settings.py
```diff
@@ -51,10 +51,12 @@
 AUTH_LDAP_CACHE_TIMEOUT = 3600
 
 AUTHENTICATION_BACKENDS = [
-    'django_auth_ldap.backend.LDAPBackend',
     "django.contrib.auth.backends.ModelBackend",
 ]
 
+if os.environ.get('LDAP_SERVER_URI'):
+	AUTHENTICATION_BACKENDS.insert(0, 'django_auth_ldap.backend.LDAPBackend')
```
mgax: It looks like flake8 really doesn't like this line:

https://jenkins.liquiddemo.org/vmck/acs-interface/753/1/2

Issue #78, opened 2019-10-22
pipenv arguments


mgax:
During vagrant provision, pipenv prints out a lot of junk:

https://jenkins.liquiddemo.org/vmck/acs-interface/757/1/3

The offending line is here:

https://github.com/vmck/acs-interface/blob/master/ci/provision-acs-interface.sh#L9

Add the `--ignore-pipfile` option so it doesn't try to generate `Pipfile.lock` again. Also, find out how to call `pipenv` so that it's quiet.

mgax:
Fixed in https://github.com/vmck/acs-interface/pull/79

PR #79, opened 2019-10-22
adding ignore-pipfile and silencing output

Zelgius:
PR for #78 

Added `--ignore-pipfile` and silenced pipenv output. 

Based on [comments on a closed bug](https://github.com/pypa/pipenv/issues/1920
) `pipenv` doesn't support a `--quite` option, so just redirecting the output to `/dev/null` will keep vagrant pretty. However failures may go unnoticed. Perhaps send output to a file instead and on successful completion delete the file? Let me know if you want me to do this. 

https://github.com/pypa/pipenv/issues/1920

### DIFF
--- a/ci/provision-acs-interface.sh
+++ b/ci/provision-acs-interface.sh
@@ -6,7 +6,7 @@ sudo pip3 install pipenv
 
 
 cd /vagrant
-pipenv install --ignore-pipfile &> /dev/null
+pipenv install --ignore-pipfile 2> /dev/null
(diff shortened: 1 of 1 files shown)


mgax:
Hey, thanks for the PR!

> Perhaps send output to a file instead and on successful completion delete the file? Let me know if you want me to do this.

Yeah, there seems to be no great way of handling this. Maybe at least redirect only `stderr` to `/dev/null`.

PR #84, opened 2019-10-25
Styling for Stdout and Stderr

Jokeswar:
Separating `stdout` and `stderr` in two tabs.


mgax:
In light of https://github.com/vmck/acs-interface/issues/95 I suggest we close this PR and revisit it if we ever want to display both streams separately.

Jokeswar:
I'll close it, and keep the branch.

PR #88, opened 2019-10-28
Fix late sub

Jokeswar:
If we have not defined holidaystart and holidayfinish `['']` is passed to the penalty function, which triggers an error.

### DIFF
--- a/interface/penalty.py
+++ b/interface/penalty.py
@@ -21,10 +21,10 @@ def compute_penalty(upload_time, deadline, penalty,
     Note: time interval between deadline and upload time (seconds)
     is time.mktime(upload_time) - time.mktime(deadline)
     """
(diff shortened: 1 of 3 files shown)


interface/penalty.py
```diff
@@ -21,10 +21,10 @@ def compute_penalty(upload_time, deadline, penalty,
     Note: time interval between deadline and upload time (seconds)
     is time.mktime(upload_time) - time.mktime(deadline)
     """
-    if holiday_start is None:
+    if holiday_start[0] == '':
```
mgax: This is a terrible way to check for an unset value. You should change the code [here](https://github.com/vmck/acs-interface/blob/0fffa6ee3480f3dd13596540b1de5da1687f28bf/interface/utils.py#L98) to set `None` instead.

Issue #91, opened 2019-10-30
Deactivate submission for expired assignments


Jokeswar:
Add another `datetime` field for a hard deadline.

Jokeswar:
Implemented in #103

PR #103, opened 2019-11-02
Hard deadline

Jokeswar:
Solves the hard deadline problem #91.


interface/models.py
```diff
@@ -43,6 +45,13 @@ class Assignment(models.Model):
     def full_code(self):
         return f'{self.course.code}-{self.code}'
 
+    @property
+    def is_active(self):
+        diff = (self.deadline_hard
+                - datetime.datetime.now(datetime.timezone.utc))
```
mgax: Please avoid this random indentation. In this case you can extract a variable:

```python
        now = datetime.datetime.now(datetime.timezone.utc)
        diff = self.deadline_hard - now
```

interface/models.py
```diff
@@ -43,6 +45,13 @@ class Assignment(models.Model):
     def full_code(self):
         return f'{self.course.code}-{self.code}'
 
+    @property
+    def is_active(self):
```
mgax: Ah, this is elegant, I like :)

interface/migrations/0009_auto_20191101_2102.py
```diff
@@ -0,0 +1,29 @@
...
+        ),
+        migrations.AddField(
+            model_name='assignment',
+            name='deadline_hard',
+            field=models.DateTimeField(null=True),
+        ),
+        migrations.AlterField(
+            model_name='submission',
+            name='assignment',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, to='interface.Assignment'),
```
mgax: I've fixed this in https://github.com/vmck/acs-interface/commit/d266ea645c948de5ebbd26c0c8af0a068b8056ae

Issue #104, opened 2019-11-04
Error downloading submissions


mgax:
The "download last submissions for review" fails with an internal error:

```
ERROR:django.request:Internal Server Error: /admin/interface/assignment/
Traceback (most recent call last):
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/contrib/admin/options.py", line 606, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/utils/decorators.py", line 142, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/contrib/admin/sites.py", line 223, in inner
    return view(request, *args, **kwargs)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/utils/decorators.py", line 45, in _wrapper
    return bound_method(*args, **kwargs)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/utils/decorators.py", line 142, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/root/.local/share/virtualenvs/interface-yvgcbbav/lib/python3.7/site-packages/django/contrib/admin/options.py", line 1698, in changelist_view
    response = self.response_action(request, queryset=cl.get_queryset(request))
[...]

PR #105, opened 2019-11-04
Skip missing files but report the error

mgax:
Fixes https://github.com/vmck/acs-interface/issues/104

### DIFF
--- a/interface/admin.py
+++ b/interface/admin.py
@@ -1,11 +1,17 @@
 from zipfile import ZipFile
 from tempfile import TemporaryDirectory
 from pathlib import Path
+import logging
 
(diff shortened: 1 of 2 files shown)


interface/admin.py
```diff
@@ -1,11 +1,17 @@
 from zipfile import ZipFile
```
Jokeswar: There is also a `submission.download()` part in [`SubmissionAdmin.download_archive()`](https://github.com/vmck/acs-interface/blob/master/interface/admin.py#L92) that could use a `try-except` block.

mgax: Good catch, fixed, thanks!

PR #106, opened 2019-11-04
Fixtures

mgax:
So that we don't click create save every time for development.


ci/fixtures.json
```diff
@@ -0,0 +1,41 @@
+[
+{
+  "model": "auth.user",
+  "pk": 1,
+  "fields": {
+    "password": "pbkdf2_sha256$150000$EoxUbjPQTZBy$OyNZocrY9aewIVmB2nbnCoZVfDh9PC9gBOtXG4Nb8t8=",
```
Jokeswar: Is this the hashed password? If so we should add a `Development` entry in the `README` where we say the actual password.

ci/fixtures.json
```diff
@@ -0,0 +1,41 @@
...
+  "model": "interface.assignment",
+  "pk": 1,
+  "fields": {
+    "course": 1,
+    "code": "a1",
+    "name": "Assignment 1",
+    "max_score": 100,
+    "deadline": "2019-12-31T22:00:00Z",
+    "repo_url": "https://github.com/vmck/assignment",
+    "repo_branch": "pc-00"
```
Jokeswar: This should get changed because #103 modified the assignment model.

Issue #92, opened 2019-10-31
Calculate final grade on submission save; disable admin editing


mgax:
The final grade is calculated based on other numbers. It should not be possible to change it from Django admin ,because it would be overwritten, thus causing confusion.

* [x] Calculate the final grade on [pre_save](https://docs.djangoproject.com/en/2.2/ref/signals/#django.db.models.signals.pre_save)
* [x] Mark the model field as read-only ([`editable=False`](https://docs.djangoproject.com/en/2.2/ref/models/fields/#editable))

PR #110, opened 2019-11-12
Add pre_save recalculation

Jokeswar:
#92 `total_score` is now read-only. On save from dajngo admin  `total_score` is recalculated. It takes into account changes from `score`, `penalty`, `review_score`. If the `review_message` is changed `review_score` will not be updated(i.e. will not be recalculated based on `review_message`), and there will be no impact on `total_score`.


Jokeswar:
Let's say a TA gives a review with the normal review button in `submission/<pk>` :
```
+1.0: great
```
This [part](https://github.com/vmck/acs-interface/blob/master/interface/views.py#L110-L121) here gets the `review_score` as `1`. Let's say that the `total_score` is 101 (100 from tests, 1 from `review_score`). Now if we change the review message to:
```
+2.0: good
```
in django admin and hit save, `total_score` will still be 101 (100 from tests, 1 from `review_score`). What I want to say is that `review_message` is not reprocessed because [this](https://github.com/vmck/acs-interface/blob/master/interface/models.py#L117-L126) only takes the `review_score` without looking at `review_message`. @mgax

mgax:
> What I want to say is that `review_message` is not reprocessed because [this](https://github.com/vmck/acs-interface/blob/master/interface/models.py#L117-L126) only takes the `review_score` without looking at `review_message`

Then why don't we move [this](https://github.com/vmck/acs-interface/blob/master/interface/views.py#L110-L121) to [here](https://github.com/vmck/acs-interface/blob/master/interface/models.py#L117-L126)?

Jokeswar:
I'll do that, and I believe we should make `review_score` read-only too.

mgax:
> I believe we should make `review_score` read-only too.

I agree.

interface/views.py
```diff
@@ -109,18 +109,8 @@ def homepage(request):
...
-        re.MULTILINE,
-    )
-    log.debug('Marks found: ' + str(marks))
-
-    review_score = sum([decimal.Decimal(mark) for mark in marks])
-
-    submission.review_score = review_score
-    submission.total_score = submission.calculate_total_score()
     submission.review_message = request.POST['review-code']
+    submission.total_score = submission.calculate_total_score()
```
mgax: Since `calculate_total_score` is called in the pre-save hook, this is a bit redundant, no?

PR #115, opened 2019-11-14
Remove the error filed

### LINKED ISSUE #95: Unify stdout and stderr as a single stream
All CI systems join the stdout and stderr streams to show a single view. Let's follow their lead. The checker and student code should have stdout and stderr set to the same file stream, so maybe append `2>&1` to `cd /vagrant; ./script.sh` [here](https://github.com/vmck/vagrant-vmck/blob/master/submission/launch.sh#L16).

Let's keep the rest of the code as-is, we can remove the `stderr` field later, after we're happy with the setup.

(Leaving the issue in the `acs-interface` repo since `vagrant-vmck` should be independent from `acs-interface`; at some point we should move that logic away from it)

gmuraru:
Fixes #95 

Remove the field with the stderr output.
This PR should be merged alongside [this](https://github.com/vmck/vagrant-vmck/pull/41)

### DIFF
--- a/interface/views.py
+++ b/interface/views.py
@@ -160,7 +160,6 @@ def done(request, pk):
     assert submission.verify_jwt(request.GET.get('token'))
 
     stdout = utils.decode(options['stdout'])
-    stderr = utils.decode(options['stderr'])
     exit_code = int(options['exit_code'])
 
     score = re.search(r'.*TOTAL: (\d+\.?\d*)/(\d+)', stdout, re.MULTILINE)
@@ -171,19 +170,14 @@ def done(request, pk):
     if len(stdout) > 32768:
         stdout = stdout[:32730] + '... TRUNCATED BECAUSE TOO BIG ...'
 
-    if len(stderr) > 32768:
-        stderr = stderr[:32730] + '... TRUNCATED BECAUSE TOO BIG ...'
-
     submission.score = decimal.Decimal(points)
     submission.penalty = submission.compute_penalty()
     submission.total_score = submission.calculate_total_score()
     submission.stdout = stdout
-    submission.stderr = stderr
     submission.state = Submission.STATE_DONE
 
     log.debug(f'Submission #{submission.id}:')
     log.debug(f'Stdout:\n{submission.stdout}')
-    log.debug(f'Stderr:\n{submission.stderr}')
     log.debug(f'Exit code:\n{exit_code}')
 
     submission.save()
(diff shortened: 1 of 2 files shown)


interface/views.py
```diff
@@ -160,7 +160,7 @@ def done(request, pk):
     assert submission.verify_jwt(request.GET.get('token'))
 
     stdout = utils.decode(options['stdout'])
-    stderr = utils.decode(options['stderr'])
+    stderr = utils.decode(options['stderr'], '')
```
mgax: You sure this works? [`decode`](https://github.com/vmck/acs-interface/blob/3683329573c9a063608f0675f5d3277449bd5dc0/interface/utils.py#L36) only accepts one argument.

interface/models.py
```diff
@@ -154,12 +154,12 @@ def compute_review_score(self):
...
-        timestamp = self.timestamp.strftime(penalty.DATE_FORMAT)
-        deadline = self.assignment.deadline_soft.strftime(penalty.DATE_FORMAT)
+        timestamp = self.timestamp or datetime.datetime.now()
+        deadline = self.assignment.deadline_soft
 
         return penalty.compute_penalty(
-            timestamp,
-            deadline,
+            timestamp.strftime(penalty.DATE_FORMAT),
+            deadline.strftime(penalty.DATE_FORMAT),
```
mgax: This bit should not be in this PR. Please `git fetch` and merge with `origin/master`.

gmuraru: Updated

interface/views.py
gmuraru: Removed the stderr from the ```views``` and ```templates```

gmuraru:
@mgax in order to test this I need [this] (https://github.com/vmck/vagrant-vmck/pull/41) merged, right?

mgax:
> @mgax in order to test this I need [this] ([vmck/vagrant-vmck#41](https://github.com/vmck/vagrant-vmck/pull/41)) merged, right?

You can make your own docker build of `vagrant-vmck`, publish it under your docker account, and configure vmck to use your image.

As a side note, vmck can be configured to use another image with `MANAGER_TAG=0.3.2`, but the repo is hardcoded. So, bonus points if you modify vmck, to allow the entire docker image name to be configured, somewhere around [here](https://github.com/vmck/vmck/blob/bc0acbd21b88c1755686b923e7b759e70296b02e/vmck/backends/submission.py#L9) :)

mgax:
Actually, you can test with the regular `vagrant-vmck` image, it will send `stderr` but that will get ignored.

PR #117, opened 2019-11-14
Windows fix

Jokeswar:
Windows has a different end of line sequence (`\r\n`) compared to Unix (`\n`) and that makes some shell scripts fail if they are run with `./` (e.g. errors like: `'/bin/bash\r' does not exist`). Now `vagrant up` should be fine on Windows.

### DIFF
--- a/ci/provision-acs-interface.sh
+++ b/ci/provision-acs-interface.sh
@@ -8,12 +8,10 @@ sudo pip3 install pipenv
 cd /vagrant
 pipenv install --ignore-pipfile 2> /dev/null
 mkdir -p data
-cp ./examples/.env .
 
 container=$(docker ps -f name=minio -aq)
 if [ -z $container ]; then (
-  sed -i 's/\r$//' examples/minio.sh
-  pipenv run bash examples/minio.sh
+  pipenv run ./examples/minio.sh
 ) fi
 
 pipenv run python manage.py migrate
--- a/ci/provision-acs-interface.sh
+++ b/ci/provision-acs-interface.sh
@@ -8,10 +8,12 @@ sudo pip3 install pipenv
 cd /vagrant
 pipenv install --ignore-pipfile 2> /dev/null
 mkdir -p data
+cp ./examples/.env .
 
 container=$(docker ps -f name=minio -aq)
 if [ -z $container ]; then (
-  pipenv run examples/minio.sh
+  sed -i 's/\r$//' examples/minio.sh
+  pipenv run bash examples/minio.sh
 ) fi
 
 pipenv run python manage.py migrate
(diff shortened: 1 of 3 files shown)


ci/provision-cluster.sh
```diff
@@ -14,7 +14,7 @@ fi
   cp examples/cluster.ini ./
   sed -i '/nomad_meta/a vmck_worker = true' cluster.ini
   sed -i '/^run_jobs =.*/c\run_jobs = dnsmasq' cluster.ini
-  bin/docker.sh --rm --pull
+  bash bin/docker.sh --rm --pull
```
gmuraru: Q: Should the ```bin/docker.sh``` file be executable? (I do not know if at the moment is)

Jokeswar: `bin/docker.sh` is executable. If I run it with bash, for files that end with `\r\n` as new line, it ignores the `\r` and the script runs as expected.

ci/provision-acs-interface.sh
```diff
@@ -8,10 +8,12 @@ sudo pip3 install pipenv
 cd /vagrant
 pipenv install --ignore-pipfile 2> /dev/null
 mkdir -p data
+cp ./examples/.env .
 
 container=$(docker ps -f name=minio -aq)
 if [ -z $container ]; then (
-  pipenv run examples/minio.sh
+  sed -i 's/\r$//' examples/minio.sh
+  pipenv run bash examples/minio.sh
```
mgax: Maybe [this SO answer](https://stackoverflow.com/a/35986153) has a more elegant fix? Ultimately, it's a bug that files end up borked inside the VM.

ci/provision-acs-interface.sh
```diff
@@ -8,10 +8,12 @@ sudo pip3 install pipenv
 cd /vagrant
 pipenv install --ignore-pipfile 2> /dev/null
 mkdir -p data
+cp ./examples/.env .
```
mgax: You  sure this is ok? It will overwrite your local `.env` file when using a shared folder. The readme now encourages the user to make their own `.env` file.

ci/provision-cluster.sh
gmuraru: And shouldn't we remove the executable right from the ```bin/docker.sh``` if we call the file like this?

mgax: Nah, leave it, it's meant to be executable, we'll probably want to call it directly sometime in the future.

ci/provision-acs-interface.sh
Jokeswar: Thanks! It looks a lot better now!

ci/provision-acs-interface.sh
Jokeswar: My bad!

PR #121, opened 2019-11-21
Submission fields

mgax:
- change stdout, stderr and review to multi-line text fields
- make more fields read-only

Fixes https://github.com/vmck/acs-interface/issues/122


interface/models.py
```diff
@@ -89,13 +89,13 @@ class Submission(models.Model):
                                    on_delete=models.CASCADE,
                                    null=True)
     user = models.ForeignKey(User, on_delete=models.PROTECT, null=True)
-    stdout = models.CharField(max_length=32768,
+    stdout = models.TextField(max_length=32768,
                               default='',
                               blank=True)
-    stderr = models.CharField(max_length=32768,
+    stderr = models.TextField(max_length=32768,
```
gmuraru: Q: Would the ```stderr`` field still be needed?

mgax: No, but we should remove it in a separate PR.

PR #128, opened 2019-11-29
Review status

### LINKED ISSUE #123: Results list - Reviewed status
If a submission is reviewed, the status filed should change from Done to Reviewed

gmuraru:
Fixes #123 
Add ```Reviewed``` column for the status field - it would check if there is any review message in the ```review``` field for a submission.

### DIFF
--- a/interface/templates/interface/submission_list.html
+++ b/interface/templates/interface/submission_list.html
@@ -19,7 +19,7 @@
             <div class="row mt-2">
                 <table class="table table-striped table-bordered table-sm">
                     <caption>Submission list</caption>
-                    <thead>
+                    <thead class="text-center">
                         <tr>
                             <th scope="col">User id</th>
                             <th scope="col">Submission id</th>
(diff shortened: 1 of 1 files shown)


interface/models.py
```diff
@@ -137,6 +137,9 @@ def update_state(self):
             self.state = response.json()['state']
             self.save()
 
+        if self.state == self.STATE_DONE and self.review_message:
+            self.state = "reviewed"
```
mgax: Currently `self.state` is a cache of the [vmck job state](https://github.com/vmck/vmck/blob/445dceae47748db454659f8d450dea63afd31f3c/vmck/models.py#L15-L19), let's leave it like that.

How about we display the review state separately? Where do you want to see it?

gmuraru: I think the best place would be at the ```Status``` field.
I am thinking to add a flag at the ```Submission``` class level and send it to the template - and depending on that flag write ```sub.status``` or ```reviewed```.

What do you think?

mgax: How about we display "done" if it's finished running, and then "done, reviewed" after it's reviewed?

PR #129, opened 2019-11-29
Audit log

Jokeswar:
I chose [easy-audit](https://github.com/soynatan/django-easy-audit) as it's quite minimal. It does the job when it comes to logging changes of objects.
REQUEST EVENTS:
image
LOGIN EVENTS:
image
CRUD(CReate Update Delete) EVENTS:
image

But it **doensn't** solve the action logging part. For that the only app that I found was [actions-logger](https://github.com/shtalinberg/django-actions-logger).  With a bit of work it can be set up, **but** from what I could learn this app it's actually a fork of https://django-auditlog.readthedocs.io/en/latest/ ( and as a bonus we barely have any docs). So two questions. Do we keep `easy-audit`? Do we use `actions-logger` or do we try to write something of our own?


mgax:
I have some reservations about the quality of this library. It has no CI, and their error handling policy is basically "if you can't log it, that's fine, move on": https://github.com/soynatan/django-easy-audit/blob/8047064b3c32964db82c733685aa3e9c66b15b2b/easyaudit/signals/auth_signals.py#L15-L16

Jokeswar:
> I've tested it locally. It logs changes made via the admin site, so it will keep a log of changes to courses and assignments, but it doesn't track e.g. reviews made by TAs. We should keep track of those too - I suspect that's where we'll need the audit log the most.

My bad. I forgot to add the admin integration. Funny part is that the changes are recorded but not displayed. Now it should show every change.

PR #134, opened 2019-12-09
Add time period between assignments

### LINKED ISSUE #96: Throttling for uploading submissions (max 5 every 5 minutes)
Let's protect against malicious students who want to fill up the submission queue. Here's a simple idea: when an upload comes in, open a transaction (https://docs.djangoproject.com/en/2.2/topics/db/transactions/#django.db.transaction.atomic), lock the assignment row (https://docs.djangoproject.com/en/2.2/ref/models/querysets/#select-for-update), and count the number of assignments for that user that fall within the 5-minute window. If they are too many then deny the upload. Otherwise proceed normally, and only release the transaction after creating the submission record. The whole thing will probably be quick enough to not cause problems.

Alternatively you can lock the user row so there will be practically no lock contention.

gmuraru:
Fixes #96 

Added a new parameter for each assignment such that we can limit the number of submissions per student


interface/backend/submission.py
```diff
@@ -1,6 +1,10 @@
 import logging
 
 import interface.backend.minio_api as storage
+from interface.models import Submission
+
+from django.db import transaction
+from django.utils import timezone
```
mgax: Please move these imports up. [As PEP8 says](https://www.python.org/dev/peps/pep-0008/#imports):

> Imports should be grouped in the following order:
> 
>     Standard library imports.
>     Related third party imports.
>     Local application/library specific imports.
> 
> You should put a blank line between each group of imports.

interface/views.py
```diff
@@ -71,7 +71,12 @@ def upload(request, course_code, assignment_code):
         form = UploadFileForm(request.POST, request.FILES)
         if form.is_valid():
             file = request.FILES['file']
-            handle_submission(file, assignment, request.user)
+            delta_t = handle_submission(file, assignment, request.user)
+
+            if delta_t:
```
mgax: Let's model this case as an exception so that it's clear this is an error situation. Define an exception in `submission.py`, e.g.

```python
class TooManySubmissionsError(Exception):
    pass
```

Throw it from `handle_submission` and catch it here.

interface/views.py
```diff
@@ -71,7 +71,12 @@ def upload(request, course_code, assignment_code):
...
+                return render(request, 'interface/upload.html',
+                              {'form': form, 'delta_t': delta_t})
```
mgax: I'm super biased against this kind of indentation. I know the codebase is full of them (hello @Jokeswar) but let's not add more. Here's a nicer way:

```python
                return render(
                    request,
                    'interface/upload.html',
                    {'form': form, 'delta_t': delta_t},
                )
```

interface/templates/interface/upload.html
```diff
@@ -13,6 +13,12 @@
         {% include "interface/tracker.html" %}
     </head>
 
+    {% if delta_t %}
+        <script type="text/javascript">
+            alert('Please wait {{delta_t}}s before doing another submission');
```
mgax: I suggest that, instead of throwing a popup in the face of the user, we show a [bootstrap alert](https://getbootstrap.com/docs/4.4/components/alerts/).

interface/templates/interface/upload.html
```diff
@@ -13,6 +13,12 @@
         {% include "interface/tracker.html" %}
     </head>
 
+    {% if delta_t %}
```
mgax: Ok, so the `if` makes sense, but it's not clear what the condition means. Let's write it as `{% if too_many_submissions %}` or something.

interface/backend/submission.py
gmuraru: Done

interface/views.py
```diff
@@ -71,7 +71,21 @@ def upload(request, course_code, assignment_code):
...
+
+            delta_t = 0
+            try:
+                handle_submission(file, assignment, request.user)
+            except TooManySubmissionsError as e:
+                delta_t = e.get_delta_t()
+                return render(
+                    request,
+                    'interface/upload.html',
+                    {
```
gmuraru: Q: Does this looks ok? (as indentations?)

mgax: Yeah that works

interface/backend/submission.py
```diff
@@ -11,10 +16,24 @@
 def handle_submission(file, assignment, user):
     log.debug(f'Submission {file.name} received')
 
-    submission = assignment.submission_set.create(
-        user=user,
-        archive_size=file.size,
-    )
+    with transaction.atomic():
+        Assignment.objects.select_for_update().get(pk=assignment.pk)
```
mgax: Please add a comment here, something like "Computing time from the last upload by this user. We lock the assignment row to prevent simultaneous uploads (race condition)."

interface/backend/submission.py
```diff
@@ -11,10 +16,24 @@
...
+        entries = Submission.objects.filter(user=user).order_by('-timestamp')
```
mgax: But since we're locking the assignment, shouldn't we be checking only uploads for this assignment?

`assignment.submission_set.filter(user=user)...`

interface/backend/submission.py
```diff
@@ -23,3 +42,11 @@ def handle_submission(file, assignment, user):
...
     log.debug(f'Submission #{submission.id} was sent to VMCK '
               f'as #{submission.vmck_job_id}')
+
+
+class TooManySubmissionsError(Exception):
+    def __init__(self, delta_t):
+        self.delta_t = delta_t
+
+    def get_delta_t(self):
+        return self.delta_t
```
mgax: It's overkill to define a getter for a field that you just know is there. Just use the attribute.

interface/views.py
```diff
@@ -71,8 +72,20 @@ def upload(request, course_code, assignment_code):
...
+            try:
+                handle_submission(file, assignment, request.user)
+            except TooManySubmissionsError as e:
+                return render(
+                    request,
+                    'interface/upload.html',
+                    {
+                        'form': form,
+                        'too_many_submissions': {'delta_t': e.get_delta_t()},
+                    }
```
mgax: Final comma pls :)

Issue #87, opened 2019-10-28
Fix submission test


mgax:
https://github.com/vmck/acs-interface/pull/86 disables the submission test because it doesn't get along with minio. Please fix :)

gmuraru:
This #136 solves this issue, right?

mgax:
> This #136 solves this issue, right?

Yep, I've linked it from the PR.

PR #136, opened 2019-12-15
Pytest

Jokeswar:
I believe we might want to break this into separate tests. Also it uses this version of `vagrant-vmck` https://github.com/vmck/vagrant-vmck/blob/vagrant-pytest/submission/launch.sh (vagrant has some trouble installing rsync in a docker container). Is there anything we might want to test?

Fixes https://github.com/vmck/acs-interface/issues/87

### DIFF
--- a/ci/test-pytest.sh
+++ b/ci/test-pytest.sh
@@ -11,5 +11,6 @@ HOSTNAME=*
 ACS_INTERFACE_ADDRESS=http://10.66.60.1:8000/
 MANAGER_TAG=vagrant-pytest" > .env
 
-exec pipenv run pytest --liveserver 10.66.60.1:8000
+pipenv run pytest --liveserver 10.66.60.1:8000
+
 rm .env
--- a/ci/test-pytest.sh
+++ b/ci/test-pytest.sh
@@ -1,4 +1,15 @@
 #!/bin/bash -ex
 
 cd /vagrant
-exec pipenv run pytest
+
+echo "MINIO_ADDRESS=10.66.60.1:9000
+MINIO_ACCESS_KEY=1234
+MINIO_SECRET_KEY=123456789
+VMCK_API_URL=http://10.66.60.1:10000/v0/
+DEBUG=True
+HOSTNAME=*
+ACS_INTERFACE_ADDRESS=http://10.66.60.1:8000/
+MANAGER_TAG=vagrant-pytest" > .env
+
+exec pipenv run pytest --liveserver 10.66.60.1:8000
+rm .env
--- a/ci/test-pytest.sh
+++ b/ci/test-pytest.sh
@@ -11,6 +11,4 @@ HOSTNAME=*
 ACS_INTERFACE_ADDRESS=http://10.66.60.1:8000/
 MANAGER_TAG=vagrant-pytest" > .env
 
-pipenv run pytest --liveserver 10.66.60.1:8000
-
-rm .env
+exec pipenv run pytest --liveserver 10.66.60.1:8000
--- a/testsuite/test_interface.py
+++ b/testsuite/test_interface.py
@@ -1,4 +1,5 @@
 import time
+from datetime import datetime
 
 import pytest
 from django.conf import settings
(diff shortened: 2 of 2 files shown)


testsuite/test_interface.py
```diff
@@ -30,16 +38,35 @@ def test_submission(client):
...
     while submission.state != submission.STATE_DONE:
-        time.sleep(2)
+        submission = Submission.objects.all()[0]
 
-    assert submission.score > 0
+        time.sleep(10)
+
+        count += 1
+        if count == 18:
+            assert 0 == 1
```
gmuraru: ```assert False```

testsuite/test_interface.py
```diff
@@ -30,16 +38,35 @@ def test_submission(client):
         )
 
     assert len(Submission.objects.all()) == 1
-    assert storage.exists('1.zip')
+    assert storage.exists('1.zip') is True
```
gmuraru: Q: Shouldn't be ok only with ```assert storage.exists(`1.zip`)```?

ci/test-pytest.sh
```diff
@@ -1,4 +1,16 @@
...
+MINIO_SECRET_KEY=123456789
+VMCK_API_URL=http://10.66.60.1:10000/v0/
+DEBUG=True
+HOSTNAME=*
+ACS_INTERFACE_ADDRESS=http://10.66.60.1:8000/
+MANAGER_TAG=vagrant-pytest" > .env
+
+pipenv run pytest --liveserver 10.66.60.1:8000
+
+rm .env
```
gmuraru: Q: ```rm -rf```?

testsuite/test_interface.py
Jokeswar: Good point. I guess I am stuck with a bit of verbosity from C

ci/test-pytest.sh
Jokeswar: Actually we might not even need this. It's only run on the CI so we don't care about clean-up.

testsuite/test_interface.py
```diff
@@ -35,11 +43,30 @@ def test_submission(client):
...
+    assert submission.score == 100
+    assert submission.total_score == 100
+
+    client.post(
+        '/submission/1/review',
+        data={'review-code': '+10.0: Good Job'},
+        HTTP_REFERER='/'
+    )
+
+    submission = Submission.objects.all()[0]
```
gmuraru: Maybe add an assert for ```objects.all()``` - that it has ```len 1```

testsuite/test_interface.py
```diff
@@ -35,11 +43,30 @@ def test_submission(client):
...
     assert submission.state == submission.STATE_NEW
-    assert submission.assignment.code == 'pc-a0'
+    assert submission.assignment.code == 'a0'
     assert submission.archive_size > 0
     assert submission.vmck_job_id > 0
 
+    count = 0
     while submission.state != submission.STATE_DONE:
-        time.sleep(2)
+        submission = Submission.objects.all()[0]
```
mgax: You can call [refresh_from_db](https://docs.djangoproject.com/en/3.0/ref/models/instances/#django.db.models.Model.refresh_from_db) to update the model instance.

testsuite/test_interface.py
```diff
@@ -35,11 +43,30 @@ def test_submission(client):
...
 
-    assert submission.score > 0
+        time.sleep(10)
```
mgax: Let's go with a shorter sleep, even 1s should be fine.

testsuite/test_interface.py
```diff
@@ -35,11 +43,30 @@ def test_submission(client):
...
+
+        count += 1
+        if count == 18:
```
mgax: I suggest you use [`time()`](https://docs.python.org/3/library/time.html#time.time) to decide when to bail from the loop, it's easier to understand when reading the code.

testsuite/test_interface.py
```diff
@@ -35,11 +43,30 @@ def test_submission(client):
...
+            assert False
+
+    assert submission.score == 100
+    assert submission.total_score == 100
+
+    client.post(
+        '/submission/1/review',
+        data={'review-code': '+10.0: Good Job'},
+        HTTP_REFERER='/'
```
mgax: Final comma pls :)

Issue #114, opened 2019-11-14
Vagrant Provisioning Fail


gmuraru:
Vagrant fails to provision with the following error:
```
==> default: Running provisioner: shell...
    default: Running: /tmp/vagrant-shell20191114-8816-1c1k12z.sh
    default: + sudo docker cp cluster:/app/bin/nomad /usr/local/bin/nomad
    default: Error: No such container:path: cluster:/app/bin/nomad
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

mgax:
Right, so this is actually a re-run of the provisioning, after the initial run failed. The first run probably got stuck [here](https://github.com/vmck/acs-interface/blob/master/ci/provision-cluster.sh#L17). A workaround is:

```shell
vagrant ssh -- rm -rf /opt/cluster
vagrant provision
```

So that it provisions the cluster again from the start.

mgax:
We could run the provision scripts as root by removing `privileged: false` from the lines in Vagrantfile. The scripts will probably need tweaking.

https://github.com/vmck/acs-interface/blob/85cf502103407f72833a8933075fab285e679d97/Vagrantfile#L13-L16

PR #138, opened 2019-12-16
Provision as root so that docker works first time

mgax:
Fixes https://github.com/vmck/acs-interface/issues/114


gmuraru:
This will remove all the warning you see when you run ```vagrant provision```?

mgax:
> This will remove all the warning you see when you run `vagrant provision`?

I'm not sure what warnings you mean. It fixes the error in the ticket.

PR #142, opened 2019-12-21
Action Logger

### LINKED ISSUE #98: Audit logging
* Log admin actions (perhaps Django already does this for us?)
* Log actions through the UI

Audit logs should answer these questions:
* Who made an action?
* When?
* What was the previous and new state of the resource that they changed?

There should be several ready-made options for us, and if we don't like them, it should be straight-forward to implement for ourselves.

Jokeswar:
Currently logs our custom actions.
Opera Snapshot_2019-12-21_135240_localhost
@mgax do we want this functionality as a separate app? Another issue is that it doesn't log `save`,`change` and `delete` actions from admin and I do not know of a way of logging them nicely (we could use signals, or overwrite `save` and `delete` and use the decorator on them). Fix #98


mgax:
Huh, the CI fails with a docker build error, which makes no sense given the changes in the PR. @Jokeswar can you please have a look?

PR #154, opened 2020-01-13
Correct redirect and sort after timestamp

gmuraru:
Fixes #152.

After uploading redirect a user to his submission page for that specific assignment.


mgax:
> Fixes #152.

I think you mean some other ticket number, that's a PR right there :)

PR #156, opened 2020-01-13
Navbar button

### LINKED ISSUE #150: Mobile menu button bug
While on a mobile browser(my case: Chrome, OS: Android), the menu button isn't responding to taps so I have to navigate using links

Jokeswar:
Fixes #150. Better order.


interface/templates/interface/generics/navbar.html
```diff
@@ -1,24 +1,34 @@
 <nav class="navbar navbar-expand-xl navbar-light bg-light">
   <a class="navbar-brand">VMCK</a>
+
   <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbarNav" aria-controls="navbarNav" aria-expanded="false" aria-label="Toggle navigation">
     <span class="navbar-toggler-icon"></span>
   </button>
+
   <div class="collapse navbar-collapse" id="navbarNav">
     <ul class="navbar-nav">
+      {% if current_user.is_authenticated %}
```
mgax: Please add some indentation for the stuff inside `if` and `else` so it's easier to figure out which is which.

interface/templates/interface/homepage.html
```diff
@@ -1,37 +1,21 @@
-<!doctype html>
-<html lang='en'>
-  <head>
-    <!-- Required meta tags -->
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+{% extends "interface/generics/base.html" %}
```
mgax: Ah, much better :)

interface/views.py
```diff
@@ -76,7 +76,10 @@ def upload(request, course_code, assignment_code):
     else:
         form = UploadFileForm()
 
-    return render(request, 'interface/upload.html', {'form': form})
+    return render(request, 'interface/upload.html', {
+        'form': form,
+        'current_user': request.user,
```
mgax: This should already be available to all templates as `user`: https://docs.djangoproject.com/en/3.0/ref/templates/api/#django-contrib-auth-context-processors-auth

interface/views.py
```diff
@@ -127,27 +131,28 @@ def submission_list(request):
...
-                   'sub_base_url': redirect(submission_list).url,
-                   'current_user': request.user,
-                   'logout_url': redirect(logout_view).url})
+    return render(request, 'interface/submission_list.html', {
+        'subs': subs,
+        'homepage_url': redirect(homepage).url,
+        'sub_base_url': redirect(submission_list).url,
+        'current_user': request.user,
+        'logout_url': redirect(logout_view).url,
+    })
```
mgax: Ok yes I really like this indentation. Thank you! :)

interface/views.py
```diff
@@ -127,27 +131,28 @@ def submission_list(request):
...
-                   'submission_list_url': redirect(submission_list).url,
-                   'fortune': fortune_msg
-                   })
+    return render(request, 'interface/submission_result.html', {
+        'sub': sub,
+        'current_user': request.user,
+        'homepage_url': redirect(homepage).url,
+        'submission_review_message': sub.review_message,
+        'submission_list_url': redirect(submission_list).url,
+        'fortune': fortune_msg
```
mgax: Missed a spot (final comma) :p

Issue #119, opened 2019-11-15
Have a separation between different courses


gmuraru:
Have a separation between course assignments.
Ex: To upload assignment for a PC homework you will need to click on PC and then select the homework

gmuraru:
For further reading: https://stackoverflow.com/a/12354293

gmuraru:
@mgax don't close this issue yet. I want to write a separate file with what features are now in vmcheckerv2

mgax:
> I want to write a separate file with what features are now in vmcheckerv2

Then why not make an issue for that? Or change this issue's title and description. I realise it means "write documentation" to you, but not to anybody else, sorry :)

And, if it's all the same to you, I suggest you do the writeup on this repo's wiki, so it's easy to update and to link.

Jokeswar:
@gmuraru is it alright if we close this? We have the wiki for the features, don't we?

gmuraru:
Yes, we have.

PR #158, opened 2020-01-21
Add separation between courses

gmuraru:
Fixes issues #119 

I left visible for a teaching assistant only the ```Submission``` and the ```Assignment``` models.


mgax:
I've merged with master and created a migration

PR #162, opened 2020-02-04
Moss

### LINKED ISSUE #140: Have MOSS integrated
[MOSS](https://theory.stanford.edu/~aiken/moss/) is a tool used for plagiarism detection - a second opinion need to be given by a human.

Integrate MOSS into vmchecker2.
Update documents on how to use it.

Jokeswar:
Integrating MOSS. Closes #140. Minor style changes and fixing a bug in `views.py/subs_for_user` where `submissions` was a tuple instead of a list.


interface/models.py
```diff
@@ -60,6 +72,11 @@ class Assignment(models.Model):
     repo_url = models.CharField(max_length=256, blank=True)
     repo_branch = models.CharField(max_length=256, blank=True)
     repo_path = models.CharField(max_length=256, blank=True)
+    language = models.CharField(
+        max_length=32,
+        choices=list(LANG_CHOICES.items()),
```
mgax: :+1: nice

interface/migrations/0019_auto_20200204_2118.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.0.2 on 2020-02-04 19:18
```
mgax: Can you please rename the migration to something descriptive like `0019_assignment_language`?

interface/templates/interface/submission_list.html
```diff
@@ -33,7 +33,7 @@
                 {% endif %}
               </td>
               <td>{{ sub.total_score }}/{{ sub.assignment.max_score }}</td>
-              <td>{{ sub.state }}</td>
+              <td>{{ sub.get_state_display }}</td>
```
mgax: Where is `get_state_display` defined? I can't find it anywhere.

interface/admin.py
```diff
@@ -133,22 +139,68 @@ def zip_submissions(self, request, submissions):
             review_zip = (tmp / 'review.zip').open('rb')
             return FileResponse(review_zip)
 
-    @log_action('Download last submissions')
-    def download_review_submissions(self, request, queryset):
+    @log_action('Run moss check')
+    def run_moss(self, request, queryset):
```
mgax: There's a lot of code in this admin view. How about you move it to another module, like `moss.py`?

interface/templates/interface/submission_list.html
Jokeswar: Got it [here](https://docs.djangoproject.com/en/3.0/ref/models/instances/#django.db.models.Model.get_FOO_display).

PR #172, opened 2020-02-27
Refactor

mgax:


interface/scoring.py
```diff
@@ -1,15 +1,39 @@
 import time
 import math
+import configparser
+import decimal
+import re
```
Jokeswar: My OCD tells me to order imports these by length.

PR #177, opened 2020-03-12
Asynchronous check for submission status

gmuraru:
Check for submissions status in an asynchronous way + add debugger logs

### DIFF
--- a/interface/management/commands/sync_vmck.py
+++ b/interface/management/commands/sync_vmck.py
@@ -1,0 +1,28 @@
+from django.core.management.base import BaseCommand
+from django.conf import settings
+
+from interface.models import Submission
+
+import time
+import logging
+
+log = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Check submission status'
+
+    def handle(self, *args, **options):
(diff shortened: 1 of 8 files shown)


interface/management/commands/sync_vmck.py
```diff
@@ -0,0 +1,28 @@
+from django.core.management.base import BaseCommand
+from django.conf import settings
+
+from interface.models import Submission
+
+import time
+import logging
```
Jokeswar: Imports from the standard library should be at the top.

runinterface
```diff
@@ -4,4 +4,5 @@ cd "$( dirname "${BASH_SOURCE[0]}" )"
 
 ./manage.py migrate
 ./manage.py storage_setup
+./manage.py sync_vmck &
```
mgax: I think it would be better to define another task (a duplicate of [this one](https://github.com/vmck/acs-deploy/blob/master/nomad_jobs/acs-interface.hcl#L116-L233)) that runs the `sync_vmck` process in a separate docker container. That way you can see the logs separately and start/stop each of them as needed.

PR #186, opened 2020-03-27
Refactor vmck options

Jokeswar:
Rewrite options sent to vmck to decouple them. Depends on vmck/vmck#152. This version is incompatible with previous versions of vmck. Tests should pass after merging vmck/vmck#152 (tests use `vmck/vmck:master` by default).

### DIFF
--- a/interface/vmck.py
+++ b/interface/vmck.py
@@ -34,21 +34,24 @@ def evaluate(submission):
     options = vmck_config(submission)
     name = f'{submission.assignment.full_code} submission #{submission.id}'
     options['name'] = name
-    options['manager'] = True
+    options['restrict_network'] = True
+
+    options['manager'] = {}
+    options['manager']['vagrant_tag'] = settings.MANAGER_TAG
+    options['manager']['memory'] = settings.MANAGER_MEMORY
+    options['manager']['cpu_mhz'] = settings.MANAGER_MHZ
+
     options['env'] = {}
-    options['env']['archive'] = submission.get_url()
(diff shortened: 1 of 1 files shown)


interface/vmck.py
```diff
@@ -34,21 +34,24 @@ def evaluate(submission):
     options = vmck_config(submission)
     name = f'{submission.assignment.full_code} submission #{submission.id}'
     options['name'] = name
-    options['manager'] = True
+    options['restrict_network'] = True
+
+    options['manager'] = {}
+    options['manager']['vagrant_tag'] = settings.MANAGER_TAG
+    options['manager']['memory'] = settings.MANAGER_MEMORY
```
gmuraru: Q: This is how much memory/cpu a virtual machine will use, right? (the machine where the student homework will run, right?)

gmuraru: Q: Maybe we should also add here, ```MB``` (to have the measurement unit as we do for the CPU?)

Jokeswar: A: Not quite. There is another task that runs alongside the student's vm (it uses [ vmck/vagrant-vmck](https://github.com/vmck/vagrant-vmck) -- the submission folder). This connects to the vm, runs the scripts and eventually sends back the results, stdout and stderr to acs-interface. These are its resources (50 Mb - of memory and 30 Mhz of cpu if i am not mistaken).

Jokeswar: You mean as a comment or change the name of the option (i. e. `memory_MB`)?

gmuraru: As the name of the option

PR #187, opened 2020-04-01
Add general queue

### LINKED ISSUE #194: General queue
Implement a general queue in Django such that there is a maximum of assignments that are sent to `Nomad`

gmuraru:
Fixes #194
This is a first implementation of the general queue mechanism


interface/models.py
```diff
@@ -166,6 +170,11 @@ def update_state(self):
         state = vmck.update(self)
         if state != self.state:
             self.state = state
+
+            if state == Submission.STATE_DONE:
+                SubmissionScheduler.show()
```
gmuraru, line 175: MAYBE: Remove this show

testsuite/test_student_workflow.py
```diff
@@ -42,9 +45,23 @@ def test_submission(client, live_server):
...
     assert submission.archive_size > 0
+
+    # As the reason mentioned above, the submission might have been already
+    # submitted
+    start = time.time()
+    while submission.vmck_job_id is None:
+        time.sleep(0.5)
+        submission.refresh_from_db()
+
+        if time.time() - start >= 2:
```
gmuraru, line 62: Add assert directly to this:
```
assert (time.time() - start) < 2
```

testsuite/test_student_workflow.py
```diff
@@ -59,14 +76,26 @@ def test_submission(client, live_server):
...
+        data={'review-code': review_message},
         HTTP_REFERER='/',
     )
 
     submission.refresh_from_db()
 
     assert len(submission.review_message) > 0
-    assert submission.review_score == 10
-    assert submission.total_score == 110
+    assert submission.review_score == 5.5
```
gmuraru, line 89: Q: Shouldn't the review score be +10? (10 - 5 + 5)

testsuite/test_student_workflow.py
```diff
@@ -59,14 +76,26 @@ def test_submission(client, live_server):
...
+    assert submission.total_score == 105.5
```
gmuraru, line 90: Q: And also here 110

testsuite/test_student_workflow.py
```diff
@@ -59,14 +76,26 @@ def test_submission(client, live_server):
...
+
+    response = client.get('/submission/1/download')
+
+    with TemporaryDirectory() as _tmp:
```
gmuraru, line 94: Q: Why not use ```tempfile```? (generate a temporary file directly)

testsuite/test_minio.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_minio(client, live_server):
+    filepath = settings.BASE_DIR / 'testsuite' / 'test.zip'
+
+    with open(filepath, 'rb') as f:
+        storage.upload('myFile.zip', bytes(f.read()))
+
+    assert storage.exists('myFile.zip')
+
+    with TemporaryDirectory() as _tmp:
```
gmuraru, line 18: Q: Why not use a temporary file instead of a temporary directory?

testsuite/test_student_workflow.py
Jokeswar: The last term is +0.5 not +5.

testsuite/test_student_workflow.py
Jokeswar: See the above

interface/backend/submission/submission_scheduler.py
```diff
@@ -0,0 +1,98 @@
...
+    general_queue = SubQueue(settings.TOTAL_MACHINES)
+
+    def __init__(self):
+        raise AttributeError("No init method for SubmissionScheduler")
+
+    @staticmethod
+    def add_submission(sub):
+        queue = SubmissionScheduler.general_queue
+
+        if queue is None:
```
gmuraru, line 85: Q: Could you remove this for now?
This is a check that does not make much sense because we have memory and I am thinking this only adds overhead

interface/backend/submission/submission.py
```diff
@@ -18,7 +19,9 @@ class CorruptZipFile(Exception):
 def handle_submission(file, assignment, user):
     log.debug(f'Submission {file.name} received')
 
-    with ZipFile(file) as archive:
+    test_file = deepcopy(file)
+
+    with ZipFile(test_file) as archive:
```
gmuraru, line 24: TODO: keep an eye after merge over this.

testsuite/test_minio.py
Jokeswar: The minio api wants a path where it will write the data. It can't use a file handle. [See this](https://github.com/vmck/acs-interface/blob/master/interface/backend/minio_api.py#L34-L41).

PR #190, opened 2020-04-03
No empty sub

### LINKED ISSUE #184: Do not allow empty submissions
If a student press the submit button without a submission archive the request will still be sent and the student will have to wait a specific period of time to submit again.

We need to cancel the request without an archive.

**Bonus**:
* Validate if archive is zip

Jokeswar:
Fixes #184. It's simply checks if the file is an actual zip and if it is not corrupt. Django takes care that the user actually uploads a non-empty archive.

### DIFF
--- a/interface/backend/submission.py
+++ b/interface/backend/submission.py
@@ -3,7 +3,6 @@ from zipfile import ZipFile
 
 from django.db import transaction
 from django.utils import timezone
-from django.contrib import messages
 
 from interface.models import Assignment
 import interface.backend.minio_api as storage
--- a/interface/backend/submission.py
+++ b/interface/backend/submission.py
@@ -1,17 +1,27 @@
 import logging
+from zipfile import ZipFile
 
 from django.db import transaction
 from django.utils import timezone
+from django.contrib import messages
 
-import interface.backend.minio_api as storage
 from interface.models import Assignment
+import interface.backend.minio_api as storage
 
 
 log = logging.getLogger(__name__)
 
 
+class CorruptZipFile(Exception):
+    pass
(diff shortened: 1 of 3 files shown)


interface/backend/submission.py
```diff
@@ -1,18 +1,27 @@
...
 
-import interface.backend.minio_api as storage
 from interface.models import Assignment
+import interface.backend.minio_api as storage
 
 
 log = logging.getLogger(__name__)
 
 
+class CorruptZipFile(Exception):
```
gmuraru, line 14: Q: Can't we use ```exception zipfile.BadZipFile```?

interface/backend/submission.py
```diff
@@ -1,18 +1,27 @@
...
+    pass
+
+
 def handle_submission(file, assignment, user):
     log.debug(f'Submission {file.name} received')
 
+    with ZipFile(file) as archive:
+        if archive.testzip():
+            raise CorruptZipFile()
```
gmuraru, line 23: Q: Can't we raise the same exception ```zipfile.BadZipFile```?

Jokeswar: A: I want to have a clear distinction between 'Hey, you sent a txt file instead of a valid zip archive' and 'Hey, you sent a zip, but it is corrupt'. Because `ZipFile(file)` raises the exception `BadZipFile` and the function `testzip` just returns the name of the first file that has a bad checksum I wanted to integrate it nicely with the already existing `try catch` clause.

gmuraru: Yep, then it is ok :)

PR #195, opened 2020-04-24
Zip.testZip() fail case

Jokeswar:
If the archive is edited manually (i.e. changing a character using a text editor) it will make testZip() raise the error `ValueError` because it is unable to read the file properly.


interface/views.py
```diff
@@ -84,6 +84,9 @@ def upload(request, course_code, assignment_code):
             except CorruptZipFile:
                 messages.error(request, 'The archive is corrupt')
 
+            except ValueError:
```
gmuraru, line 87: Why not do multiple exceptions catch?

PR #199, opened 2020-05-09
Move submission actions

Jokeswar:
Move individual submission actions from admin to the submission page.


interface/actions_logger.py
```diff
@@ -3,7 +3,16 @@
...
+def log_action(msg, usr, obj):
+    ActionLog.objects.create(
+        timestamp=timezone.now(),
+        user=usr,
+        action=msg,
+        content_object=obj,
+    )
+
+
+def log_action_admin(msg):
```
gmuraru, line 15: Q: Should we move this to the admin file?

interface/admin.py
```diff
@@ -51,6 +51,7 @@ def _add_new_ta(self, user):
                 Permission.objects
                 .filter(codename__in=CourseAdmin._ta_permissions)
             )
+        print(user)
```
gmuraru, line 54: TODO: Remove this

interface/views.py
```diff
@@ -130,10 +135,51 @@ def homepage(request):
...
+            not in submission.assignment.course.teaching_assistants.all()):
+        return Http404('You are not allowed!')
+
     submission.review_message = request.POST['review-code']
     submission.changeReason = 'Review'
     submission.save()
 
+    log_action("Review submission", request.user, submission)
+
+    return redirect(request.META['HTTP_REFERER'])
```
gmuraru, line 148: MAYBE: Change this line with:
```
return HttpResponseRedirect(request.META.get('HTTP_REFERER', '/'))
```

interface/views.py
```diff
@@ -130,10 +135,51 @@ def homepage(request):
...
+    if (request.user
+            not in submission.assignment.course.teaching_assistants.all()):
+        return Http404('You are not allowed!')
+
+    submission.state = Submission.STATE_NEW
+    submission.evaluate()
+
+    log_action("Rerun submission", request.user, submission)
+
+    return redirect(request.META['HTTP_REFERER'])
```
gmuraru, line 165: MAYBE: Here also

PR #207, opened 2020-05-29
Code coverage

Jokeswar:
Integrate codecov.io to better gauge the efficiency of the testsuite.


ci/test-pytest.sh
```diff
@@ -2,13 +2,9 @@
...
-MINIO_ACCESS_KEY=1234
-MINIO_SECRET_KEY=123456789
-VMCK_API_URL=http://10.66.60.1:10000/v0/
-DEBUG=True
-HOSTNAME=*
-ACS_INTERFACE_ADDRESS=http://10.66.60.1:8000/
-MANAGER_TAG=vagrant-pytest" > .env
+# set the necessary envs if they are not present
+if [ ! -f .env ]; then
+    cp ./examples/.env .
```
gmuraru, line 7: Q: Why can this happen?

Jokeswar: If a developer wants to run the testsuite and an `.env` already exists we will overwrite it, and this might break testing if he set the environment differently. Also it is easier to set up the `.env` file for testing in the CI.

PR #213, opened 2020-05-31
Student tests

Jokeswar:

PR #214, opened 2020-06-01
Test Views

Jokeswar:
Test that views return the correct template. Blocked by #213.


testsuite/test_views.py
```diff
@@ -0,0 +1,151 @@
...
+    STC.assertContains(response, course.name)
+    STC.assertContains(response, assignment.name)
+
+
+@pytest.mark.django_db
+def test_submission_list(client, STC, base_db_setup):
+    (_, _, user, course, assignment) = base_db_setup
+    client.login(username=user.username, password='pw')
+
+    for i in range(30):
```
gmuraru, line 46: Can you put ```_``` for ```i```?

PR #220, opened 2020-06-12
Mock config file

### LINKED ISSUE #218: Mock config file
We should mock the config file during tests so that we don't have external dependencies. Helps #211.

Jokeswar:
Fix #218


testsuite/test_student_workflow.py
```diff
@@ -32,7 +68,7 @@ def test_submission(client, live_server, base_db_setup):
             format='multipart',
         )
 
-    assert len(Submission.objects.all()) == 1
+    assert Submission.objects.all().count() == 1
```
gmuraru, line 71: Q: Could you update in test suite all the stuff that is using ```len``` and not ```count```

testsuite/test_student_workflow.py
```diff
@@ -1,29 +1,66 @@
...
+
+    class Server:
+        def __init__(self):
+            self.server = HTTPServer(
+                (ADDR, PORT),
+                SimpleHTTPRequestHandler,
+            )
+
+        def start_server(self):
+            thread = threading.Thread(target=self.server.serve_forever)
```
gmuraru, line 35: MAYBE: Add ```daemon=True``` as a parameter to the Thread

PR #226, opened 2020-06-19
Add generic evaluator

### LINKED ISSUE #225: Generic submission evaluator
Have a generic interface through which acs-interface should send its submissions. The API should include:
 * `evaluate` -- takes a submission and evaluates it
 * `update` -- should return the state of the given submission(Running, Done...)

gmuraru:
Fixes #225

Create an abstract class for the ```evaluator``` and extend the ```vmck```


interface/backend/submission/evaluator/vmck.py
```diff
@@ -0,0 +1,71 @@
+import configparser
+from urllib.parse import urljoin
+import logging
+
+from django.conf import settings
+import requests
```
Jokeswar, line 6: Please keep the order:
```suggestion
import logging
import configparser
from urllib.parse import urljoin

import requests
from django.conf import settings
```

PR #230, opened 2020-06-26
Limit filesize for upload

### LINKED ISSUE #227: Limit file upload size
We should limit the student's archive size to stop DOS type attacks.

Jokeswar:
Fix #227.

### DIFF
--- a/interface/views.py
+++ b/interface/views.py
@@ -14,6 +14,7 @@ from django.contrib import messages
 from django.core.paginator import Paginator
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST
+from django.template.defaultfilters import filesizeformat
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth import authenticate, login, logout
 from django.http import JsonResponse, FileResponse, Http404
@@ -92,9 +93,13 @@ def upload(request, course_pk, assignment_pk):
             else:
                 return redirect(users_list, course_pk, assignment_pk)
 
-            return render(request, 'interface/upload.html', {
-                'form': form,
-            })
+        elif request.FILES.get('file_size_warning', None):
+            msg = (
+                f"Filesize is bigger than "
+                f"{filesizeformat(settings.FILE_UPLOAD_MAX_MEMORY_SIZE)}"
+            )
+            messages.error(request, msg)
+
     else:
         form = UploadFileForm()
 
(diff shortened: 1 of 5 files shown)


testsuite/test_student_workflow.py
```diff
@@ -198,3 +200,28 @@ def test_anonymous(client, STC):
...
+    zip_archive = zipfile.ZipFile(buff, mode='w')
+    zip_archive.writestr('test.c', 'aaaa'*2**20)
+
+    upload = SimpleUploadedFile(
+        FILEPATH.name,
+        buff.read(),
+        content_type='application/zip',
+    )
+
+    client.post(
```
gmuraru, line 221: Q: You said in the call that we can not see the error message right?
Thinking if we can see the error message on the webpage? (too big archive kind of stuff)

Jokeswar: The problem that I encountered is that I do not know, at the moment, how to talk from the upload handler to the rest of the code. Looks like I do not have a way to catch the exception `StopUpload`. I intend to look a bit into `handle_raw_input`, the docs say that i can return a tuple of `(POST, FILES)`, it might help.

interface/views.py
```diff
@@ -92,9 +93,13 @@ def upload(request, course_pk, assignment_pk):
             else:
                 return redirect(users_list, course_pk, assignment_pk)
 
-            return render(request, 'interface/upload.html', {
-                'form': form,
-            })
+        elif request.FILES.get('file_size_warning', None):
```
gmuraru, line 96: Q: Isn't something like this cleaner? (you might have investigated this)
https://stackoverflow.com/questions/38842053/how-to-check-the-file-size-limit-in-django

gmuraru: Thinking that we can inherit the UploadFileForm and add that function

Jokeswar: The process of uploading a file is something similar to:

1. Client selects the file
2. The upload handlers save the file on our system
3. The form data validation

The resource you pointed to, it's cleaner, but the check happens on the third stage where we already have taken the time and resources to get the entire file on our server. Maybe I can clear this mess by returning a fake `UploadedFile` with no data, but with the size of infinity so that we can use the restriction you pointed to. I will look into it!

Jokeswar:
@gmuraru I believe it should be cleaner now

gmuraru:
LGTM

PR #239, opened 2020-07-26
View for teaching assistants to navigate through source code 

### LINKED ISSUE #163: Create a View where teaching assistants can navigate through the source code
Currently, a teaching assistant can view the source code for an assignment only by downloading and opening the source files using an editor.

We want to have a GitHub like structure for each submission where the teaching assistant can view the source files from that submission.

Depends on #243

BiancaNecula:
Fixes #163 

The ta can see the source code or other files in submission directly in browser by accessing /submission/<id_of_submission>/<filename>.


interface/views.py
```diff
@@ -327,3 +328,21 @@ def reveal(request, course_pk, assignment_pk):
...
     ))
+
+
+@login_required
+@staff_member_required
+def code_view(request, pk, filename):
+    submission = get_object_or_404(Submission, pk=pk)
+
+    if (request.user
+            not in submission.assignment.course.teaching_assistants.all()):
```
gmuraru, line 339: Q: Might be worth creating a variable for this:
Ex:
```all_tas = submission.assignment.course.teaching_assistants.all()```
and then check:
```
if request_user not in all_tas:
```
Such that we do not split the if statement in two lines

File: interface/backend/minio_api.py [hunk 2]
```diff
@@ -31,6 +31,15 @@ def copy(source_file, destination_file):
                         f'/{settings.MINIO_BUCKET}/{source_file}')
 
 
+def download_buffer(filename, buffer):
+    try:
+        data = _client.get_object(settings.MINIO_BUCKET, filename)
+    except NoSuchKey:
+        raise MissingFile(filename)
+    for chunck in data.stream(32*1024):
```
gmuraru, line 39: Q: Can we create a global variable for this value? (with uppercase letters)

interface/codeview.py
```diff
@@ -0,0 +1,26 @@
+import io
+import logging
+from zipfile import ZipFile
+
+from interface.backend.minio_api import MissingFile, download_buffer
+
+log = logging.getLogger(__name__)
+
+
+def submission_cut(request, submission, filename):
```
gmuraru, line 10: Q: From what comes the name ```submission_cut```?

interface/codeview.py
```diff
@@ -0,0 +1,26 @@
...
+    try:
+        download_buffer(f'{submission.pk}.zip', buff)
+    except MissingFile:
+        return io.StringIO("The archive is missing!")
+
+    submission_archive = ZipFile(buff)
+    try:
+        file = submission_archive.read(
+            filename,
+            pwd=None,
```
Jokeswar, line 21: `pwd` is, by default, set to `None`. No need to set it explicitly.

interface/codeview.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+    except MissingFile:
```
Jokeswar, line 23: I don't think `zipfile.read()` throws this kind of error. Let's leave it as generic as possible and remove `MissingFile` for the moment.

interface/views.py
```diff
@@ -327,3 +328,21 @@ def reveal(request, course_pk, assignment_pk):
...
+        return HttpResponse(status=403)
+
+    file = submission_cut(request, submission, filename)
+
+    file_content = file.read()
+
+    file.close()
+    context = {'file_content': file_content}
```
Jokeswar, line 347: Let's simplify this to something less spaced out. What do you say about:
```
context = {'file_content': file.read()}
file.close()
```

interface/codeview.py
BiancaNecula: It was the preliminary form of the function. It's actually `extract_file`.

interface/backend/minio_api.py
```diff
@@ -7,6 +7,8 @@
 
 import io
 
+BLOCK_SIZE = 32*1024
```
Jokeswar, line 10: Let's have this in settings.py

testsuite/test_student_workflow.py
```diff
@@ -344,3 +344,30 @@ def test_user_see_revealed_score(client, STC, base_db_setup):
...
+            format='multipart',
+        )
+
+    submission = Submission.objects.all()[0]
+
+    response = client.get(
+        f'/submission/{submission.pk}/test',
+        follow=True,
+    )
+    STC.assertTemplateNotUsed(response, 'interface/code_view.html')
```
Jokeswar, line 373: You could add a check for the response status and if it redirects to the admin page.

testsuite/test_student_workflow.py
```diff
@@ -344,3 +344,30 @@ def test_user_see_revealed_score(client, STC, base_db_setup):
...
+    FILEPATH = settings.BASE_DIR / 'testsuite' / 'test.zip'
+
+    (_, _, user, course, assignment) = base_db_setup
+
+    client.login(username=user.username, password='pw')
+
+    with open(FILEPATH, 'rb') as file:
+        upload = SimpleUploadedFile(FILEPATH.name,
+                                    file.read(),
+                                    content_type='application/zip')
```
Jokeswar, line 360: Let's go with a uniform coding style:
```
                                   # code ....
                                   content_type='application/zip',
)
```

testsuite/test_ta_workflow.py
```diff
@@ -795,3 +798,31 @@ def test_ta_see_revealed_score(client, STC, base_db_setup):
...
+    FILEPATH = settings.BASE_DIR / 'testsuite' / 'test.zip'
+
+    (_, ta, _, course, assignment) = base_db_setup
+
+    client.login(username=ta.username, password='pw')
+
+    with open(FILEPATH, 'rb') as file:
+        upload = SimpleUploadedFile(FILEPATH.name,
+                                    file.read(),
+                                    content_type='application/zip')
```
Jokeswar, line 814: Same here as above

interface/codeview.py
```diff
@@ -0,0 +1,23 @@
...
+    buff = io.BytesIO()
+    try:
+        download_buffer(f'{submission.pk}.zip', buff)
+    except MissingFile:
+        return io.StringIO("The archive is missing!")
+
+    submission_archive = ZipFile(buff)
+    try:
+        file = submission_archive.read(filename)
+    except Exception:
```
Jokeswar, line 20: The actual exception that we are looking for is `KeyError`. `Exception` is the father of all exceptions and would catch more than necessary.

interface/views.py
```diff
@@ -327,3 +328,20 @@ def reveal(request, course_pk, assignment_pk):
...
     ))
+
+
+@login_required
+@staff_member_required
+def code_view(request, pk, filename):
+    submission = get_object_or_404(Submission, pk=pk)
+    all_tas = submission.assignment.course.teaching_assistants.all()
+
+    if (request.user not in all_tas):
```
gmuraru, line 339: MAYBE: Remove parentheses

testsuite/test_student_workflow.py
```diff
@@ -344,3 +344,38 @@ def test_user_see_revealed_score(client, STC, base_db_setup):
     )
     STC.assertNotContains(response, "N/A")
     STC.assertContains(response, "Final score")
+
+
+@pytest.mark.django_db
+def test_user_code_view(client, STC, base_db_setup):
+    FILEPATH = settings.BASE_DIR / 'testsuite' / 'test.zip'
```
gmuraru, line 351: Q: Could you remove this? it is already defined at the top of this file

testsuite/test_ta_workflow.py
```diff
@@ -795,3 +798,33 @@ def test_ta_see_revealed_score(client, STC, base_db_setup):
 
     response = client.get(f'/assignment/{course.pk}/{assignment.pk}')
     STC.assertNotContains(response, "N/A")
+
+
+@pytest.mark.django_db
+def test_ta_code_view(client, STC, base_db_setup):
+    FILEPATH = settings.BASE_DIR / 'testsuite' / 'test.zip'
```
gmuraru, line 805: Also here

PR #255, opened 2020-09-05
Add unify config

### LINKED ISSUE #249: Unify assignment options
## Description
We should move all the options present in `config.ini` on acs-interface to centralize them.

## Is your feature request related to a problem?
If we want to benchmark assignments we need to keep our settings together.

gmuraru:
## Description
Remove the config file for assignments and add it to the ```Assignment``` model - in this way we unify all the configurations.

Fixes #249 

## Testing
* Testing

## Checklist
- [X] I have labeled this PR with the relevant [Type labels](https://github.com/vmck/acs-interface/labels)
- [X] Tests (if they apply)


interface/backend/submission/evaluator/vmck.py
```diff
@@ -34,7 +35,7 @@ def evaluate(submission):
             f"token={str(submission.generate_jwt(), encoding='latin1')}"
         )
 
-        options = VMCK._vmck_config(submission)
+        options = deepcopy(submission.assignment.options)
```
gmuraru, line 38: Need to add more changes here

Jokeswar (approved):

PR #256, opened 2020-09-18
Comment section for code review

### LINKED ISSUE #165: Comment and grade directly on the code
It depends on #163.
Having the possibility to view the source code directly in the browser, the teaching assistant should now be able to add comments on specific lines (specify cases of bad indentation, parts of the code which they liked, etc.) from the code.
By adding comments like:
```
+1.0: Nice idea
```
The student should receive 1.0 bonus points to the grade.
```
-1.0: Did not check the returned value
````
The student should receive a 1.0 penalty to the grade.

### PR DESCRIPTION
**Pull request by BiancaNecula**
## Description
<TODO> - This feature allows the ta to review and grade the homework directly on the code. They can leave multiple comments on every line (which will be highlighted) and change the score. The student can reply back.

Fixes #165 

## Testing
- To do
## Checklist
- [ ] Tests (if they apply)

### CODE REVIEW

File: interface/scoring.py [hunk 1]
```diff
@@ -94,9 +94,27 @@ def compute_review_score(submission):
     return sum([decimal.Decimal(mark) for mark in marks])
 
 
+def compute_comments_review(submission):
+    total_sum = 0
+    for comment in submission.comments.all():
+        if (
+            comment.user
+            in submission.assignment.course.teaching_assistants.all()
```
gmuraru, line 102: Q: Could you save this variable before the outer for? and then do only
```
if comment.user in teaching_assistants:
```

interface/templates/interface/code_view_homepage.html
```diff
@@ -0,0 +1,39 @@
...
+            {% with dict=tree.items template="interface/tree_view.html" %}
+                {% include template %}
+            {% endwith %}
+        </div>
+        <div class="col-lg-9">
+            <div class="row align-items-center">
+                <div class="col-lg-2"></div>
+                <div class="col-lg-9">
+                    <pre>
+                        /^--^\     /^--^\     /^--^\
```
gmuraru, line 17:
```
 .       .
 |\_---_/|
/   o_o   \
|    U    |
\  ._I_.  /
 `-_____-'
```

testsuite/test_student_workflow.py
```diff
@@ -163,11 +163,11 @@ def test_user_cannot_review(client, STC, base_db_setup):
 
     review_message = "+10.0: Hacker"
     response = client.post(
```
gmuraru, line 165: Q: Could you also put a test where the student comments with something like? :pray: 
```
+10.0: I really did a nice job!
```

interface/views.py
```diff
@@ -350,14 +349,58 @@ def code_view(request, pk, filename):
...
+        "file_content": file.getvalue(),
+        "tree": tree,
+        "pk": submission.pk,
+        "sub": submission,
+        "path": filename,
+        "file_exists": False,
+    }
+
+    if (
+        file.getvalue() != "The file is missing!"
```
gmuraru, line 378: Q: Could you save that value to don't having to call the same method twice?
like
```
file_value = file.getValue()
```
and then you can do
```
if file_value not in ["The file is missing!", "The archive is missing!"]:
```

interface/views.py
```diff
@@ -350,14 +349,58 @@ def code_view(request, pk, filename):
     submission = get_object_or_404(Submission, pk=pk)
     all_tas = submission.assignment.course.teaching_assistants.all()
 
-    if request.user not in all_tas:
+    if request.user not in all_tas and request.user is not submission.user:
         return HttpResponse(status=403)
 
     tree = tree_view(request, submission)
 
+    if filename == "code":
```
Jokeswar, line 357: This might introduce the issue of actually having a file named `code` in the archive. A solution could be having two paths in our `urlpatterns`:
```
"submission/<int:pk>/code/",
"submission/<int:pk>/code/<path:filename>/",
```
Those two should point to different views. One will handle the if case and the other the else case. It will also make it clearer where you are in your logic (empty page / page with code).

interface/views.py
```diff
@@ -350,14 +349,58 @@ def code_view(request, pk, filename):
...
+        comment = None
+        if request.method == "POST":
+            form = CommentForm(request.POST)
+            if form.is_valid():
+                comment = form.save(commit=False)
+                comment.submission = submission
+                comment.user = request.user
+                comment.path = filename
+                comment.line = request.POST.get("line", "")
+                comment.save()
```
Jokeswar, line 392: Maybe you can add hidden rows in your form so those can be filled in by form.save().

Jokeswar (changes requested):
Overall :+1:

interface/views.py
```diff
@@ -405,3 +394,23 @@ def code_view(request, pk, filename):
...
+
+
+@login_required
+@staff_member_required
+def code_view_homepage(request, pk):
+    submission = get_object_or_404(Submission, pk=pk)
+    all_tas = submission.assignment.course.teaching_assistants.all()
+
+    if request.user not in all_tas and request.user is not submission.user:
+        return HttpResponse(status=403)
```
gmuraru, line 406: Q: Could you also add a test for this? :+1:

gmuraru (approved):
Great job!

PR #268, opened 2020-10-03
Recomputes total score when the soft deadline changes

### LINKED ISSUE #181: Recompute total score if TAs change the deadline
The penalty deadline for submissions should be recomputed, if the deadline changes, without user intervention. Currently if TAs change the deadline they have to select all affected submissions and select the `Recompute score` action.

clovis1122:
## Description
Adds a change on the admin site to recompute the submissions when the soft deadline changes.

Fixes #181 

## Testing
Check the included test.

## Checklist
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/vmck/acs-interface/labels)
- [x] Tests (if they apply)

### DIFF
--- a/interface/models.py
+++ b/interface/models.py
@@ -94,7 +94,11 @@ class Assignment(models.Model):
         branch = self.repo_branch or "master"
         path = f"{self.repo_path}/" if self.repo_path else ""
         return urljoin(url_base, f"{branch}/{path}{filename}")
-
+    
+    def refresh_submission_penalty(self):
+        for submission in self.submission_set.all():
+            submission.penalty = None
+            submission.save()
 
 class Submission(models.Model):
     """ Model for a homework submission
--- a/testsuite/conftest.py
+++ b/testsuite/conftest.py
@@ -1,12 +1,16 @@
 from datetime import datetime, timezone
+from tempfile import NamedTemporaryFile
+from http.server import HTTPServer, SimpleHTTPRequestHandler
+import threading
 import pytest
+
 
 from django.test import SimpleTestCase
 from django.http import JsonResponse
 from django.contrib.auth.models import User
 from django.contrib.admin.sites import AdminSite
 
-from interface.models import Course, Submission
+from interface.models import Course, Submission, Assignment
 from interface.admin import CourseAdmin, AssignmentAdmin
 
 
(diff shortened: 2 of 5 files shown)


interface/models.py
```diff
@@ -94,7 +94,11 @@ def url_for(self, filename):
         branch = self.repo_branch or "master"
         path = f"{self.repo_path}/" if self.repo_path else ""
         return urljoin(url_base, f"{branch}/{path}{filename}")
-
+    
+    def refresh_submission_penalty(self):
+        for submission in self.submission_set.all():
+            submission.penalty = None
+            submission.save()
```
clovis1122, line 101: I'm using it this way (instead of calling a single update) to trigger the signal that updates the submission.

testsuite/conftest.py
```diff
@@ -74,3 +78,30 @@ def download_all_submissions_stub(assignment, request, queryset):
         "download_all_submissions",
         download_all_submissions_stub,
     )
+
+
+@pytest.fixture
+def mock_config(monkeypatch):
```
clovis1122, line 84: Moved this fixture here, since I need it for the TA test.

testsuite/test_ta_workflow.py
```diff
@@ -275,6 +275,44 @@ def test_ta_edit_assignment(STC, client, base_db_setup):
...
+def test_soft_deadline_change_trigger_recompute(client, base_db_setup, mock_config):
+    (_, ta, user, course, assignment) = base_db_setup
+
+    mock_config.start_server()
+
+    submission = assignment.submission_set.create(
+        score=100.00, state=Submission.STATE_DONE, user=user,
+    )
+
+    assert submission.penalty == 0
```
clovis1122, line 288: Probably not needed? Let me know.
```suggestion
```

testsuite/test_ta_workflow.py
```diff
@@ -275,6 +275,44 @@ def test_ta_edit_assignment(STC, client, base_db_setup):
...
+
+    client.login(username=ta.username, password="pw")
+
+    deadline_soft = datetime.now() - timedelta(days=365)
```
clovis1122, line 292: To avoid dealing with holidays, I just used a full year and applied the maximum penalty.

testsuite/test_ta_workflow.py
gmuraru: I am ok with removing it, @Jokeswar what do you think?

testsuite/test_ta_workflow.py
```diff
@@ -275,6 +275,44 @@ def test_ta_edit_assignment(STC, client, base_db_setup):
...
+    }
+
+    response = client.post(
+        f"/admin/interface/assignment/{assignment.pk}/change/",
+        data=assignment_change_params,
+    )
+    mock_config.stop_server()
+
+    submission.refresh_from_db()
+    assert submission.penalty == 7
```
gmuraru, line 313: Q: Could you also check the score? It should be 0 I think

gmuraru:
I redirected this to the dev branch

testsuite/test_ta_workflow.py
clovis1122: So the `score` is 100, but the `total_score` is 93 (score - penalty). I think we can assert the following:

```suggestion
    assert submission.penalty == 7
    assert submission.total_score == submission.score - submission.penalty 
```

...or

```suggestion
    assert submission.penalty == 7
    assert submission.total_score != submission.score
```

testsuite/conftest.py
Jokeswar: Great :+1:. I forgot about moving it.

testsuite/test_ta_workflow.py
Jokeswar: The first option is suitable, but you could also check that total_score is 93
```suggestion
    assert submission.penalty == 7
    assert submission.total_score == submission.score - submission.penalty 
    assert submission.total_score == 93
```

clovis1122:
Done! Let me know if there's anything else needed before merging.

gmuraru (approved):
Everything looks good to me :D! Great job!

PR #269, opened 2020-10-04
Replace asserts

### LINKED ISSUE #263: Create exceptions and replace all asserts
## Description
Currently, we use asserts for things like input/response validation. Those asserts would be compiled out if you want to generate optimized bytecode.

Replace all asserts with custom made ```Exceptions``` (those exceptions should be placed in a single file in our repo).

## Expected Behavior
Have a file with all the custom made exceptions and all asserts replaced with them

clovis1122:
## Description
Replaces `asserts` statements and instead raises exceptions. I only found one assertion that is not inside the `testsuite` folder.

Follow up: should I create a custom `JwtException` class? I didn't do it since its just one...

Fixes #263

## Testing
-

## Checklist
- [x] Tests (if they apply)

### DIFF
--- a/testsuite/test_views.py
+++ b/testsuite/test_views.py
@@ -6,6 +6,7 @@ from django.conf import settings
 from interface.models import User, Submission
 from interface import utils
 
+
 def test_login(client, STC):
     response = client.get("/")
 
@@ -29,7 +30,7 @@ def test_homepage(client, STC, base_db_setup):
     (_, _, user, course, assignment) = base_db_setup
     client.login(username=user.username, password="pw")
 
-    response = client.get(f"/homepage/")
+    response = client.get("/homepage/")
 
     STC.assertTemplateUsed(response, "interface/homepage.html")
     assert response.context["courses"]
@@ -46,7 +47,9 @@ def test_submission_list(client, STC, base_db_setup):
 
     for _ in range(30):
         assignment.submission_set.create(
-            user=user, score=100.00, state=Submission.STATE_DONE,
+            user=user,
+            score=100.00,
+            state=Submission.STATE_DONE,
         )
 
     response = client.get("/submission/")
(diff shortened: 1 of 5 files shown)


gmuraru:
Aren't there more "asserts" or is only this one?

clovis1122:
Only one I found @gmuraru, the rest are on the testsuite folder.

image

gmuraru:
I think it is ok like this. I was expecting that it might be more and making custom exceptions for those might make sense - for now it is ok :+1: 
Could you also add a quick test where you call done and you expect and exception?

clovis1122:
Hey @gmuraru, so while creating the test, I realized returning an HTTP 400 response is more appropriate in this case, to let the one calling the API know that there's a mistake with the token. I added a happy path test, as well as testing this HTTP 400 response. Let me know if this is ok.

gmuraru:
Yep, from my side, it looks great! Waiting for @Jokeswar to give the second approval.

testsuite/test_views.py
```diff
@@ -73,6 +74,49 @@ def test_submission_result_done(client, STC, base_db_setup):
...
+    (_, _, user, course, assignment) = base_db_setup
+    client.login(username=user.username, password="pw")
+
+    submission = assignment.submission_set.create(
+        user=user,
+        score=0,
+        state=Submission.STATE_RUNNING,
+        stdout="Runnning",
+    )
+    token = 'yabayaba'
```
gmuraru, line 109: Love this! From the Flintstones Family, right?

gmuraru (approved):

Jokeswar:
@clovis1122 if you could run [black formatter](https://github.com/psf/black) and [flake8](https://pypi.org/project/flake8/) before we merge and fix any errors that flake8 might report, it would be great.

gmuraru:
@clovis1122 is black formatter or flake8 reporting any problems?

clovis1122:
I'll try running it later today. Sorry for the delay!

gmuraru:
No problem, thank you for the contribution :D

clovis1122:
Done, please check the last commit  @Jokeswar @gmuraru

gmuraru:
LGTM! @Jokeswar what do you think?

Jokeswar (approved):

Issue #265, opened 2020-10-03
Check if valid json before returning the state


gmuraru:
## Description
If we have a server error in ```vmck``` we might not get back a valid json in [here](https://github.com/vmck/acs-interface/blob/master/interface/backend/submission/evaluator/vmck.py#L67-L69) and we get an exception.

Introduce a new state for a submission, like ```ERR```, which means that there was an error while trying to run the submission.
If the json is not valid, return this new state.

## Expected Behavior
If we have an error on the server-side, we should indicate that in the homework.

GuillaumeFalourd:
Hi @gmuraru , would something like this be enough ?

```def update(submission):
       try: 
           response = requests.get(
                urljoin(
                    settings.VMCK_API_URL, f"jobs/{submission.evaluator_job_id}"
                )
            )
       except requests.exceptions.HTTPError as errh:
            return response.json()["HTTP ERR"]
       except requests.exceptions.ConnectionError as errc:
           return response.json()["CONNECTING ERR"]
       except requests.exceptions.Timeout as errt:
           return response.json()["TIMEOUT ERR"]
       except requests.exceptions.RequestException as err:
          return response.json()["ERR"]
       
       return response.json()["state"]
```

Or did you imagined something different ?

gmuraru:
Hmm...this would not work ```response.json()``` since it has thrown an error.
You should do:
```
  try: 
           response = requests.get(
                urljoin(
                    settings.VMCK_API_URL, f"jobs/{submission.evaluator_job_id}"
                )
            )
       except Exception e:
            return "Err" <-- and here this value should be mapped with a state from Submission

       return response.json()["state"]
```

GuillaumeFalourd:
`return "Err" <-- and here this value should be mapped with a state from Submission`
This new STATE should be mapped here : https://github.com/vmck/acs-interface/blob/3cc73d28a4a7f92cd3dd35b264fbe7e7da3acd07/interface/models.py#L115 ?

clovis1122:
Is this issue available to be worked on?

gmuraru:
@clovis1122 I think @GuillaumeFalourd might want to work on it. If not, you can take it.

gmuraru:
> `return "Err" <-- and here this value should be mapped with a state from Submission`
> 
> Should this new STATE be mapped here ?
> 
[...]

PR #270, opened 2020-10-04
New state ERROR for submission update method

GuillaumeFalourd:
## Description

- Introduce a new state error for a submission.
- If the json is not valid during the submission update, the method will now return the new state.

Fixes https://github.com/vmck/acs-interface/issues/265

## Testing
- Validated through the /test_internals.py class

## Checklist
- [ ] I have labeled this PR with the relevant [Type labels](https://github.com/vmck/acs-interface/labels) --> I don't have access
- [x] Tests

### DIFF
--- a/interface/backend/submission/evaluator/vmck.py
+++ b/interface/backend/submission/evaluator/vmck.py
@@ -75,4 +75,9 @@ class VMCK(Evaluator):
             log.debug(f"Submission #{submission.pk} update error: {e}")
             return "Error"
         else:
-            return response.json()["state"]
+            try:
+                json_response = response.json()
+                return json_response["state"]
+            except Exception as e:
+                log.debug("JSON conversion error", e)
+                return "Error"
--- a/interface/backend/submission/evaluator/vmck.py
+++ b/interface/backend/submission/evaluator/vmck.py
@@ -71,7 +71,8 @@ class VMCK(Evaluator):
                     settings.VMCK_API_URL, f"jobs/{submission.evaluator_job_id}"
                 )
             )
-        except:
+        except Exception as e:
+            log.debug(f"Submission #{submission.pk} update error: {e}")
             return "Error"
         else:
             return response.json()["state"]
--- a/interface/backend/submission/evaluator/vmck.py
+++ b/interface/backend/submission/evaluator/vmck.py
@@ -65,10 +65,13 @@ class VMCK(Evaluator):
         return response.json()["id"]
 
     def update(submission):
-        response = requests.get(
-            urljoin(
-                settings.VMCK_API_URL, f"jobs/{submission.evaluator_job_id}"
+        try:
+           response = requests.get(
+                urljoin(
+                    settings.VMCK_API_URL, f"jobs/{submission.evaluator_job_id}"
+                )
             )
-        )
-
-        return response.json()["state"]
+        except:
+            return "Error"
+        else:
+            return response.json()["state"]
--- a/interface/backend/submission/evaluator/vmck.py
+++ b/interface/backend/submission/evaluator/vmck.py
@@ -65,19 +65,14 @@ class VMCK(Evaluator):
         return response.json()["id"]
 
     def update(submission):
-        try:
-           response = requests.get(
-                urljoin(
-                    settings.VMCK_API_URL, f"jobs/{submission.evaluator_job_id}"
-                )
+        response = requests.get(
+            urljoin(
+                settings.VMCK_API_URL, f"jobs/{submission.evaluator_job_id}"
             )
+        )
+        try:
+            json_response = response.json()
+            return json_response["state"]
         except Exception as e:
-            log.debug(f"Submission #{submission.pk} update error: {e}")
+            log.debug("JSON conversion error", e)
             return "Error"
-        else:
-            try:
-                json_response = response.json()
-                return json_response["state"]
-            except Exception as e:
-                log.debug("JSON conversion error", e)
-                return "Error"
--- a/interface/backend/submission/evaluator/vmck.py
+++ b/interface/backend/submission/evaluator/vmck.py
@@ -74,5 +74,5 @@ class VMCK(Evaluator):
             json_response = response.json()
             return json_response["state"]
         except Exception as e:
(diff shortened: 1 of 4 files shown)


GuillaumeFalourd:
I was wondering where would be the best location to add tests for this feature @gmuraru ?

I was thinking of implementing a new test on the test_views.py class creating a submission with the `Submission.STATE_ERROR` state, but I don't think it will be enough 🤔

gmuraru:
Hmm...I think for this we would have to monkey patch the update function to return an error and then check the submission state.
What do you think @GuillaumeFalourd @Jokeswar ?

GuillaumeFalourd:
You tell me if it's worth it, or if it would more interesting to create another ISSUE for this part :octocat:

gmuraru:
I think they can go into the same issue since when you add a functionality ideally you would also add a test alongside it to test it :D

GuillaumeFalourd:
No problem with me. I'm new with python so it will be my first opportunity to implement test methods 🔝 
What would be the best way to do it according to the project structure ?

Jokeswar:
If we want to have a test that checks this we could create a submission (by hand with `Submission.objects.create()`) and have its `evaluator_job_id` set to a value that does not exists in the database (something like 9999 should work). Also we should not have the submission in the update queue anymore (we should change `submission.update_state()` to exit if the state is `Error`).

GuillaumeFalourd:
This test should be implemented in the `test_views.py` class @Jokeswar ?

Jokeswar:
@GuillaumeFalourd a new file might work better, `test_internals.py` as this test is quite unrelated to the others.

GuillaumeFalourd:
Ok, will try it here. Thank you 👍

GuillaumeFalourd:
I implemented the test and successfully ran it, but I just don't get why my `submission` object hasn't been updated after calling the `Submission.update_state(submission)` method.

I might be doing something wrong... Any tips ?

Screenshot 2020-10-04 at 18 39 47

interface/backend/submission/evaluator/vmck.py
```diff
@@ -65,10 +65,14 @@ def evaluate(submission):
         return response.json()["id"]
 
     def update(submission):
-        response = requests.get(
-            urljoin(
-                settings.VMCK_API_URL, f"jobs/{submission.evaluator_job_id}"
+        try:
```
gmuraru, line 68: Hmm...the exception is triggered (I think) when you try to convert to the json.
So you should have something like

```
try:
    json_response = response.json()
    return json_response["state"]
except Excetpion as e:
    log.debug(...)
    return "Error"

**Comment by GuillaumeFalourd** [comment]
I push the code for you to see, but I don't think the problem is there @gmuraru , as even with your suggestion I still don't get any return value from the `Submission.update_state(submission)` I call in the unit test.

File: interface/backend/submission/evaluator/vmck.py [hunk 2]
```diff
@@ -75,4 +75,9 @@ def update(submission):
             log.debug(f"Submission #{submission.pk} update error: {e}")
             return "Error"
         else:
-            return response.json()["state"]
\ No newline at end of file
+            try:
```
gmuraru, line 78: I think you can remove the ```try catch``` from line 68

GuillaumeFalourd: Done.
Unfortunately the error on the unit test remains the same.
Isn't the problem on the test method ?

GuillaumeFalourd:
Done @gmuraru 👍 
Unfortunately, the error on the unit test remains the same.
Isn't the problem on the test method ?

interface/models.py
```diff
@@ -152,7 +154,7 @@ class Submission(models.Model):
     history = HistoricalRecords()
 
     def update_state(self):
-        if self.state == self.STATE_DONE or self.evaluator_job_id is None:
+        if self.state == self.STATE_DONE or self.evaluator_job_id is None or self.STATE_ERROR:
```
gmuraru, line 157: What you can do here is the following.
``` if self.state in [self.STATE_DONE, self.STATE_ERROR] or ...```
Hmm...could you check if adding this makes it working?

Comment (GuillaumeFalourd) [reply interface/models.py:157]: You nailed it, this solved the problem @gmuraru !

Screenshot 2020-10-04 at 21 18 30

Didn't imagine a double `or` in a if statement could lead to this in Python 🤔

**Comment by GuillaumeFalourd** [comment]
Now that it worked @gmuraru , do you know which `evaluator_job_id` value would return the `STATE_DONE` ? So I could checked all scenarios for the `update_state` method ? I didn't find the `evaluator_job_id` values list associated for each submission state.

**Comment by Jokeswar** [comment]
@GuillaumeFalourd there is no need for that. It is already tested. `evaluator_job_id` is the id that the evaluator assigns to a job request (similar to how a submission gets its id aka pk, when a user submits its homework).

**Comment by GuillaumeFalourd** [comment]
Ok, I believed there were some values already configured for integration tests, to check specific workflows.
(I wanted to test all the `update_state` method scenarios)

I submit the tests. Let me know if I should update something.

File: interface/models.py [hunk 4]
```diff
@@ -152,7 +154,7 @@ class Submission(models.Model):
     history = HistoricalRecords()
 
     def update_state(self):
-        if self.state == self.STATE_DONE or self.evaluator_job_id is None:
+        if self.state in [self.STATE_DONE, self.STATE_ERROR] or self.evaluator_job_id is None :
             return
 
         state = SubmissionScheduler.evaluator.update(self)
```
gmuraru, line 160: Q: @Jokeswar should we also update on line 164?
If we have
```if state in {Submission.STATE_DONE, Submission.STATE_ERROR}```

Jokeswar: You are right. We will get our semaphore locked otherwise.

GuillaumeFalourd: Done it @gmuraru and @Jokeswar 👍

gmuraru (approved):
LGTM!

gmuraru:
Thank you for your contribution! :+1:

GuillaumeFalourd:
Thank you for your patience explaining everything to me @gmuraru and @Jokeswar 👍

PR #273, opened 2020-10-10
Preliminary score with penalty

### LINKED ISSUE #257: Preliminary score should take into account the penalty for being late
## Description
Preliminary score should take into account the penalty for being late as to not confuse students. The reveal should be only for preliminary score + review score.

BiancaNecula:
## Description
The preliminary score is calculated with penalty for being late. 
Total score is now just preliminary score +/- review score.

Fixes #257 

## Testing
- UI

## Checklist
- [ ] Tests (if they apply)


interface/models.py
```diff
@@ -146,6 +146,9 @@ class Submission(models.Model):
     )
     score = models.DecimalField(max_digits=5, decimal_places=2, null=True)
     penalty = models.DecimalField(max_digits=5, decimal_places=2, null=True)
+    preliminary_score = models.DecimalField(
+        max_digits=5, decimal_places=2, null=True
+    )
```
Jokeswar, line 151: There is no need to save it as a field. Usually, in a database we do not write redundant data  - `total_score` is written the same way and we should change it. We could just use `calculate_preliminary_score` with a `@property` annotation (so that it can be used in a template), provided we move it in the submission model. Maybe we could also add caching (we have cachetools) - with a TTL  type cache.

Jokeswar (changes requested):

PR #302, opened 2021-03-07
#297

### LINKED ISSUE #297: Suggestive error message for incorrect review syntax
## Description
500 Internal Server Error is received when the teaching assistant saves a review with incorrect syntax

## How to Reproduce
1. Go to Assignment Review dialog
2. Try to save the review with a first line beginning with:
-0..0: test
3. See error - 500 Server Error

## Expected Behavior
An error message should be sent back to the teaching assistant telling him that the format for the Review is wrong, instead of receiving a 500 error

Lucian-MihaiStan:
## Description
Suggestive error message for incorrect review syntax
Regex fixed for "-0..0: test" and normal tests
Link to syntax format added in "Review" block

Fixes #297 (link the issue)

## Testing

## Checklist
- [x] Tests (if they apply)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
Jokeswar, line 1: This comes from VSCode config files. They are not needed in the commit as it "forces" an editor on the developers. Also they can change from one machine to another. I recommend you use a [.gitignore_global](https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) to automatically omit the .vscode folder when doing a commit. [Git ignore rules](https://www.atlassian.com/git/tutorials/saving-changes/gitignore).

interface/templates/interface/generics/review_modal.html
```diff
@@ -11,7 +11,14 @@ <h5 class="modal-title" id="reviewLabel">Review submission {{ sub.pk }}</h5>
         <form id="reviewForm" method="post" action="review">
           {% csrf_token %}
           <div class="form-group">
-            <label for="message-text" class="col-form-label">Review details:</label>
+            <div class="row">
+                <div class="col">
+                  <label for="message-text">Review details:</label>
+                </div>
+                <div class="col">
+                  <a href="https://github.com/vmck/acs-interface/wiki/Features#review-submissions" class="float-right" style="text-decoration: none;">Format syntax</a>
```
Jokeswar, line 19: Can you please supply and image so we can see the result better :smile:.

interface/scoring.py
```diff
@@ -80,7 +80,7 @@ def compute_penalty(
 
 def compute_review_score(submission):
     marks = re.findall(
-        r"^([+-]\d+\.*\d*):",
+        r"^([+-]\d+(?:\.\d+)?):",
```
Jokeswar, line 83: :+1: nice one with the no capture group.

Jokeswar (changes requested):
We are off to a good start :+1:. As a side note, can you please add a more descriptive title to your pull request. I know what the issue #297 is about, but in the future I might not be aware of the issue that the pull request is trying to solve, so a good title might help :smile:.

Lucian-MihaiStan (commented):
I will try to add an image/icon so the modications will be much more visible.

interface/templates/interface/generics/review_modal.html
Lucian-MihaiStan: cropped

PR #304, opened 2021-03-08
Suggestive error format review

Lucian-MihaiStan:
## Description
SuggestiveError FormatReview

Fixes #302  (link the issue)

## Testing

## Checklist
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/vmck/acs-interface/labels)
- [x] Tests (if they apply)


.gitignore
```diff
@@ -3,3 +3,4 @@
 /interface/static/
 /.vagrant
 *.pyc
+/.vscode/
```
Jokeswar, line 6: The .gitignore of the repo is still not a good place. If someone for whatever reason uses intellij, they will have a .idea folder that must be ignored. This will pollute the `.gitignore` file again. The [resource](https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) I send you previously stated that you can run the following command:
```
$ git config --global core.excludesfile ~/.gitignore_global
``` 
Now any ignore rules that you add to the `.gitginore_global` will apply to all repos.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
Jokeswar, line 1: You still need to remove this file and then commit the change. `.gitignore` works on a "from here on we do this..." basis, meaning it will prevent future files that are ignored from being added to the commit. It will **not** remove already committed ones.

Jokeswar (changes requested):
Just a little push and it will be ready.

PR #305, opened 2021-03-10
Button student allow up

### LINKED ISSUE #275: Add button for allowing students to upload
## Description
When a teaching assistant makes an assignment the students are allowed to upload homework immediately.
We want to change that behavior such that we would allow students to upload only when the TA has a flag activated. 

## Is your feature request related to a problem?
Allow only TAs to upload and test homework.

Lucian-MihaiStan:
## Description
Add button for allowing students to upload

Fixes #275 

## Testing

## Checklist
- [x] Tests (if they apply)


interface/templates/interface/homepage.html
```diff
@@ -18,7 +18,7 @@
       {% for assignment in assignments %}
         <li>
           {{ assignment.name }}:
-          {% if assignment.is_active %}
+          {% if assignment.is_active and assignment.assignment_on or assignment.is_active and user in assignment.course.teaching_assistants.all %}
```
Jokeswar, line 21: It might not be enough just to hide the upload button. Students could still find the url to the upload page for this assignment and use it. A check in the upload view is still necessary

Jokeswar (changes requested):

